PR #13, opened 2022-03-03
Refactoring the whole codebase.

hjpotter92:
Adding support for gpg verification.


README.md
```diff
@@ -1,22 +1,95 @@
...
-You'll presently need the following repos cloned into the same directory as livepeer-in-a-box:
+You'll presently need the following repos cloned into the same
+directory as livepeer-in-a-box:
 
--   [go-livepeer](https://github.com/livepeer/go-livepeer) (any branch, only there for `install_ffmpeg.sh`)
--   [mistserver](https://github.com/DDVTECH/mistserver) (`livepeer-in-a-box` branch)
--   [task-runner](https://github.com/livepeer/task-runner) (whichever branch you want to use)
+ - [go-livepeer](https://github.com/livepeer/go-livepeer) (any branch, only there for `install_ffmpeg.sh`)
+ - [mistserver](https://github.com/DDVTECH/mistserver) (`livepeer-in-a-box` branch)
+ - [task-runner](https://github.com/livepeer/task-runner) (whichever branch you want to use)
```
iameli, line 10: Is this still true? Don't we have automated task-runner builds?

cmd/downloader/downloader.go
```diff
@@ -218,74 +105,57 @@ func ExtractZipArchive(archiveFile, extractPath string, service Service) {
...
 
-func ExtractTarGzipArchive(archiveFile, extractPath string, service Service) {
+// ExtractTarGzipArchive processes a tarball file and extracts a
+// single file from the service definition.
+func ExtractTarGzipArchive(archiveFile, extractPath string, service types.Service) error {
+	var outputPath string = ""
 	file, _ := os.Open(archiveFile)
 	archive, err := gzip.NewReader(file)
-	CheckError(err)
+	utils.CheckError(err)
```
iameli, line 117: If we're now following Go best practices and returning an `error` instead of just panicing all the time, we might as well do it for all errors, yeah?

cmd/downloader/downloader.go
```diff
@@ -218,74 +105,57 @@ func ExtractZipArchive(archiveFile, extractPath string, service Service) {
...
+	cliFlags := cli.GetCliFlags(buildFlags)
 	var waitGroup sync.WaitGroup
 	manifest := ParseYamlManifest(cliFlags.ManifestFile)
 	for _, element := range manifest.Box {
 		if element.Skip {
 			continue
 		}
 		waitGroup.Add(1)
-		DownloadService(cliFlags, manifest, element, &waitGroup)
+		go DownloadService(cliFlags, manifest, element, &waitGroup)
```
iameli, line 158: Is there a reason to pass in `waitGroup` as an argument instead of something like...
```suggestion
		go func(element Service) {
			DownloadService(cliFlags, manifest, element)
			waitGroup.Done()
		}(element)
```

hjpotter92: ah yes! closures. this is indeed better!

PR #17, opened 2022-04-11
workflow: Build mistserver only on tag push or manifest changes

hjpotter92:
Update pattern for when the release and codesign actions take place


hjpotter92:
weird that it picked the wrong commit message as PR title

iameli:
no longer necessary

PR #23, opened 2022-04-29
Catalyst with mapic

iameli:
Needed for the next generation of our servers. I dunno. Maybe we just include it in all the builds.


iameli:
old and bad

PR #30, opened 2022-07-05
merge catalyst-node repo

iameli:


cmd/catalyst-node/catalyst-node.go
```diff
@@ -0,0 +1,284 @@
...
+	"github.com/peterbourgon/ff/v3"
+)
+
+type (
+	catalystConfig struct {
+		serfRPCAddress           string
+		serfRPCAuthKey           string
+		mistLoadBalancerEndpoint string
+	}
+)
```
leszko, line 27:
```suggestion
type catalystConfig struct {
	serfRPCAddress           string
	serfRPCAuthKey           string
	mistLoadBalancerEndpoint string
}
```

cmd/catalyst-node/catalyst-node.go
```diff
@@ -0,0 +1,284 @@
...
+	defer client.Close()
+
+	eventCh := make(chan map[string]interface{}, 1024)
+	streamHandle, err := client.Stream("*", eventCh)
+	if err != nil {
+		return fmt.Errorf("error starting stream: %w", err)
+	}
+	defer client.Stop(streamHandle)
+
+	// eli note: not sure if this is useful yet, but we can do this as well:
```
leszko, line 66: How about removing the commented-out code?

cmd/catalyst-node/catalyst-node.go
```diff
@@ -0,0 +1,284 @@
...
+	flag.CommandLine.Parse(nil)
+
+	config := catalystConfig{
+		serfRPCAddress:           *serfRPCAddress,
+		serfRPCAuthKey:           *serfRPCAuthKey,
+		mistLoadBalancerEndpoint: *mistLoadBalancerEndpoint,
+	}
+
+	go func() {
+		// eli note: i put this in a loop in case client boots before server.
```
leszko, line 259: How about using a waitgroup (or `struct{}` channel) instead of this command and `time.Sleep()`?

cmd/catalyst-node/catalyst-node.go
```diff
@@ -0,0 +1,284 @@
...
+func init() {
+	ui := &cli.BasicUi{Writer: os.Stdout}
+
+	// eli note: this is copied from here:
+	// https://github.com/hashicorp/serf/blob/a2bba5676d6e37953715ea10e583843793a0c507/cmd/serf/commands.go#L20-L25
+	// but we should someday get a little bit smarter and invoke serf directly
+	// instead of wrapping their CLI helper
+
+	Commands = map[string]cli.CommandFactory{
+		"agent": func() (cli.Command, error) {
```
leszko, line 40: What is the meaning of `agent` in the catalyst's nomenclature?

cmd/catalyst-node/catalyst-node.go
```diff
@@ -0,0 +1,284 @@
...
+func runClient(config catalystConfig) error {
+
+	client, err := connectSerfAgent(config.serfRPCAddress, config.serfRPCAuthKey)
+
+	if err != nil {
+		return err
+	}
+	defer client.Close()
+
+	eventCh := make(chan map[string]interface{}, 1024)
```
leszko, line 59: Just wonder, what is the reason of `1024` as the channel size?

cmd/catalyst-node/catalyst-node.go
```diff
@@ -0,0 +1,284 @@
...
+	// 	return fmt.Errorf("error starting monitor: %s", err)
+	// }
+	// defer client.Stop(monHandle)
+
+	// eli note: uncertain how we handle dis/reconnects here. but it's local, so hopefully rare?
+	event := <-eventCh
+	inbox := make(chan map[string]interface{}, 1)
+	go func() {
+		for {
+			e := <-eventCh
```
leszko, line 80: What is the logic here? I can't grasp it?

leszko (changes requested):
Added some comments. But in general, I cannot grasp the logic behind this change. Do we have it explained somewhere? Looks like a lot of temp stuff, but I don't see the big picture, so I might be missing something.

cmd/catalyst-node/catalyst-node.go
iameli: No reason, we should probably make it unbuffered

cmd/catalyst-node/catalyst-node.go
iameli: Yeah, we had a whole Discord discussion about this pattern - take a look here, and then at the thread under Yondon's post https://discord.com/channels/423160867534929930/445623681251672064/986702686407761920

cmd/catalyst-node/catalyst-node.go
gioelecerati: The `agent` in the main serf process which handle the events and holds membership information.  
More here: https://www.serf.io/docs/agent/basics.html

PR #44, opened 2022-07-13
Add Mist LoadBalancer and E2E Tests

leszko:
Add Mist LoadBalancer and E2E Tests.


test/e2e/e2e_test.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func startCatalyst(t *testing.T, ctx context.Context, hostname string, network string) *catalystContainer {
+	configAbsPath, err := filepath.Abs("../../config")
+	require.NoError(t, err)
+
+	var req = testcontainers.ContainerRequest{
+		Image:        "livepeerci/catalyst:pr-42",
+		ExposedPorts: []string{tcp(webConsolePort), tcp(serfPort), tcp(httpPort), tcp(rtmpPort)},
+		ShmSize:      1874000000,
```
hjpotter92, line 37: Should it be hardcoded value? This is decided by the shared memory buffer system has and could/should be computed at runtime

leszko: yeah, I just took it from the Docker commands sent by @iameli . I think we can tweak it, decrease, or calculate at runtime.

iameli (approved):
Great work! Can easily see how this can be extended to a sophisticated and maintainable set of integration tests.

On my local machine, `make test` works about half the time. When it fails, I get:

```
--- FAIL: TestMultiNodeCatalyst (310.07s)
    e2e_test.go:179:
        	Error Trace:	e2e_test.go:179
        	            				e2e_test.go:57
        	Error:      	Condition never satisfied
        	Test:       	TestMultiNodeCatalyst
FAIL
FAIL	command-line-arguments	310.228s
FAIL
make: *** [test] Error 1
```

How shall we go about diagnosing this kind of problem? Log file dump somewhere maybe?

test/e2e/e2e_test.go
leszko: Removed `ShmSize`

leszko:
> Great work! Can easily see how this can be extended to a sophisticated and maintainable set of integration tests.
> 
> On my local machine, `make test` works about half the time. When it fails, I get:
> 
> ```
> --- FAIL: TestMultiNodeCatalyst (310.07s)
>     e2e_test.go:179:
>         	Error Trace:	e2e_test.go:179
>         	            				e2e_test.go:57
>         	Error:      	Condition never satisfied
>         	Test:       	TestMultiNodeCatalyst
> FAIL
> FAIL	command-line-arguments	310.228s
> FAIL
> make: *** [test] Error 1
> ```
> 
> How shall we go about diagnosing this kind of problem? Log file dump somewhere maybe?

Two comments on that:

1. I redirected container's logs into stdout + added `-v --logtostderr` to `go test`, so now you can see all the logs while running `make test`.
2. I checked why the test is flaky and actually it's not the test that is flaky, but the `livepeer:catalyst` image. Try running the following command a few times and you should see that it hangs sometimes. Probably there is some race condition in the catalyst itself which we should track and address separately.
```
docker run   --shm-size=1874m   --rm   -it   -v $(pwd)/config:/config:ro   --hostname catalyst-one   --name catalyst-one   --network catalyst-nodes  -p 1936:1935 -p 4242:4242   -p 7373:7373 -p 8080:8080 livepeer/catalyst MistController -c /config/catalyst-one.json
```

PR #47, opened 2022-07-16
test dtscs

iameli:
TODO before merging:

* Get TLS certs in CI
* Dynamic Mist config file generation so we can have one one DTSC test and one DTSCS test

### DIFF
--- a/cmd/catalyst-node/catalyst-node.go
+++ b/cmd/catalyst-node/catalyst-node.go
@@ -242,6 +242,8 @@ func getMistLoadBalancerServers(endpoint string) (map[string]interface{}, error)
 func execBalancer(balancerArgs []string) error {
 	cmd := exec.Command("MistUtilLoad", balancerArgs...)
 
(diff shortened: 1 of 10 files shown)


cmd/catalyst-node/catalyst-node.go
```diff
@@ -272,7 +274,7 @@ func main() {
 	mistLoadBalancerEndpoint := fs.String("mist-load-balancer-endpoint", "http://127.0.0.1:8042/", "Mist util load endpoint")
 	version := fs.Bool("version", false, "Print out the version")
 	runBalancer := fs.Bool("run-balancer", true, "run MistUtilLoad")
-	balancerArgs := fs.String("balancer-args", "", "arguments passed to MistUtilLoad")
+	balancerArgs := fs.String("balancer-args", "-S", "arguments passed to MistUtilLoad")
```
leszko, line 277: What is this `-S` parameter? Could you maybe update the description?

PR #51, opened 2022-07-21
changes for multi-node launch

iameli:


cmd/catalyst-node/catalyst-node.go
```diff
@@ -334,9 +314,34 @@ func main() {
 		}
 	}()
 
+	// Everything past this is booting up Serf
+	tmpFile, err := writeSerfConfig(&serfConfig)
```
leszko, line 318: Do we need to write the config into the file? Can't serf read the config directly from string or config flags?

cmd/catalyst-node/catalyst-node.go
```diff
@@ -334,9 +314,34 @@ func main() {
...
+	defer os.Remove(tmpFile)
+	glog.V(6).Infof("Wrote serf config to %s", tmpFile)
+
+	ui := &cli.BasicUi{Writer: os.Stdout}
+
+	// eli note: this is copied from here:
+	// https://github.com/hashicorp/serf/blob/a2bba5676d6e37953715ea10e583843793a0c507/cmd/serf/commands.go#L20-L25
+	// but we should someday get a little bit smarter and invoke serf directly
+	// instead of wrapping their CLI helper
+
```
leszko, line 331:
```suggestion
	// copied from:
	// https://github.com/hashicorp/serf/blob/a2bba5676d6e37953715ea10e583843793a0c507/cmd/serf/commands.go#L20-L25
	// we should consider invoking serf directly instead of wrapping their CLI helper
```
Nit: I suggest to not `eli note` in the comments.

cmd/catalyst-node/catalyst-node.go
```diff
@@ -348,3 +353,34 @@ func main() {
 
 	os.Exit(exitCode)
 }
+
+func writeSerfConfig(config *agent.Config) (string, error) {
+	// Two steps to properly serialize this as JSON: https://github.com/spf13/viper/issues/816#issuecomment-1149732004
```
leszko, line 358: Why do we need this hack/workaround? Can't we just create a type struct with `json` tags and marshal it?

leszko (commented):
Added some comments.

cmd/catalyst-node/catalyst-node.go
iameli: We totally could, there just already exists such a struct so I figured this would both save time and increase our confidence that we had used the correct JSON fields for Serf. A single-line ugly hack seems potentially worth it in that context?

cmd/catalyst-node/catalyst-node.go
iameli: Config file seemed expedient, especially when combined with the configuration `agent.Config` struct I'm using from the library. We'll change this up when we're creating a serf instance directly, of course

.github/workflows/catalyst.yaml
```diff
@@ -9,9 +9,6 @@ on:
       - "Dockerfile"
       - "docker-compose.yml"
       - "Makefile"
-    branches:
-      - "main"
-      - "*catalyst*"
```
iameli, line 14: @hjpotter92 Any harm in this?

cmd/catalyst-node/catalyst-node.go
leszko: I'd probably favor having slightly more code, but no hacks ;) Anyway, up to you. Btw. you can check something similar I did here: https://github.com/livepeer/catalyst/pull/53/files#diff-a904e00073eb9c69508bd6a1c0ba249d3f7dc854094a8e6380db1bd9077c888cR9

.github/workflows/catalyst.yaml
hjpotter92: i'd leave `main` here, and remove only the `*catalyst*` match. the builds would be triggered for all pull request, the additional `branches` filter applies to direct push to these branches (so we can still have branches which don't necessarily need these builds triggered).

PR #54, opened 2022-07-21
e2e_test: integrate MistLoadTest to verify streaming

emranemran:
Use MistLoadTest to spawn multiple clients for a specific protocol (e.g. HLS) as viewers at catalyst-two node. The test passes if every client is able to connect and stream successfully. The json files generated by MistLoadTest are parse to verify if streams passed or failed.

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -2,14 +2,17 @@ package e2e
 
 import (
 	"context"
+	"encoding/json"
 	"flag"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"math/rand"
 	"net/http"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
@@ -104,6 +107,7 @@ func TestMultiNodeCatalyst(t *testing.T) {
 	defer p.Kill()
 
 	requireReplicatedStream(t, c2)
+	requireProtocolLoad(t, c2)
 	requireStreamRedirection(t, c1, c2)
 }
 
@@ -293,9 +297,11 @@ func requireReplicatedStream(t *testing.T, c *catalystContainer) {
 		content := string(body)
 		for _, expected := range []string{"RESOLUTION=1920x1080", "FRAME-RATE=30", "index.m3u8"} {
 			if !strings.Contains(content, expected) {
+				glog.Info("Failed to get HLS manifest")
 				return false
 			}
 		}
+		glog.Info("Got HLS manifest!")
 		return true
 	}
 	require.Eventually(t, correctStream, 5*time.Minute, time.Second)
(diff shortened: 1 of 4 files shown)


test/e2e/e2e_test.go
```diff
@@ -207,10 +220,82 @@ func requireReplicatedStream(t *testing.T, c1 *catalystContainer, c2 *catalystCo
...
+
+	}
+
+	// Parse output results for each protocol tested above
+	var p bool = true
+	for prot := range protocols {
+
+		matches, err := filepath.Glob("*" + strings.ToUpper(prot) + "*json")
+		if err != nil {
+			panic(fmt.Errorf("Glob failed: %w", err))
```
iameli, line 266: Will defer to @leszko here on best practices for tests like this, but should this maybe be returning an error instead of panicking? Seems like a good practice in general... on the other hand I can see how this error would be like "test harness failure" instead of "failing test" and maybe a panic makes sense there.

test/e2e/e2e_test.go
```diff
@@ -207,10 +220,82 @@ func requireReplicatedStream(t *testing.T, c1 *catalystContainer, c2 *catalystCo
...
+		}
+		fmt.Println(matches)
+
+		content, err := ioutil.ReadFile(matches[0])
+		if err != nil {
+			glog.Fatalf("Error while opening output file: %s", err)
```
iameli, line 272: In any event, no sense in doing anything else after a `Fatalf`, which will crash the process immediately

test/e2e/e2e_test.go
```diff
@@ -207,10 +220,82 @@ func requireReplicatedStream(t *testing.T, c1 *catalystContainer, c2 *catalystCo
...
 				return false
 			}
 		}
+		glog.Info("Got HLS manifest!")
 		return true
 	}
 	require.Eventually(t, correctStream, 5*time.Minute, time.Second)
+
+	// Define a set of protocols to test at output of node catalyst-two
+	protocols := map[string]string{
```
iameli, line 233: Nice! I'm thinking that each protocol should probably be its own test, rather than just having one big loop in here? Minimally, having a separate function for running a protocol test will let you `return err` in the error cases for those instead of having to do awkward `continue`s or `break`s

iameli (changes requested):
Really really nice! Happy it wasn't too terrible to get this working properly - just suggested a few structuring changes

test/e2e/e2e_test.go
emranemran: Yeah, I'm inclined to remove the panic and just make it a general "fail". We'd be reviewing logs anyways to see where the failures are - doesn't seem like a panic would be anymore useful here.

test/e2e/e2e_test.go
emranemran: Ah I realize that now after looking at the `Fatalf` man page which calls os.Exit right after. Will fix.

test/e2e/e2e_test.go
emranemran: Right, I think I got a little lost in trying to do things in the "idiomatic" go style. I'll take a look at refactoring.

leszko: Yeah, it should be a separate test. Also, the very idiomatic in Go is to use a tests [] array. Please check "Introducing table driven tests" in https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

leszko (changes requested):
Added some comments. But two general thoughts:
1. We should create more tests rather than expanding the assertion of the same test. Having more tests give us better information what failed. One single test is hard to analyze in case of failures.
2. I'd try to not cleanup files manually, but rather depend on the built-in 1`t.TempDir()`

test/e2e/e2e_test.go
```diff
@@ -207,10 +220,82 @@ func requireReplicatedStream(t *testing.T, c1 *catalystContainer, c2 *catalystCo
...
+
+		cmdMistLoadTest := exec.Command("../../bin/./MistLoadTest", url)
+		out, err := cmdMistLoadTest.Output()
+		require.NoError(t, err)
+		if err != nil {
+			glog.Fatalf("Failed to start MistLoadTest: %s", err)
+		}
+		glog.Infof("MistLoadTest stdout: %s", out)
+
+		defer cmdMistLoadTest.Process.Kill()
```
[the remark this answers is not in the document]
emranemran: `MistLoadTest` will timeout and exit within 30s (default). I'll turn it into a parameter in the next push and verify it does work as expected.

test/e2e/e2e_test.go
```diff
@@ -207,10 +220,82 @@ func requireReplicatedStream(t *testing.T, c1 *catalystContainer, c2 *catalystCo
...
 	require.Eventually(t, correctStream, 5*time.Minute, time.Second)
+
+	// Define a set of protocols to test at output of node catalyst-two
+	protocols := map[string]string{
+		"hls": fmt.Sprintf("http://localhost:%s/hls/stream+foo/index.m3u8", c2.http),
+		// add more protocol <-> url mapping here to test in the future
+	}
+
+	// rm any old output files from previous MistLoadTest runs
+	cleanupFiles("*json")
```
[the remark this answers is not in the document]
emranemran: Yeah, I'll have to update `MistLoadUtil` to accept a different path. But agree managing files is not the best way - didn't know about `t.TempDir()` so i'll try to get that to work.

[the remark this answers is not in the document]
emranemran: PR here to allow output directory: 
https://github.com/DDVTECH/mistserver/pull/58/commits/1341ffbcf865814f7ebbe2bf67632177987f7146

This should let me refactor into something much simpler.

emranemran:
**Updates:** 
* Refactored to use table driven test as suggested by @leszko. We can now run a load test against each protocol (e.g. HLS) independently or all protocols at the same time as defined in the test matrix
* Using t.TempDir() to get simplify code that tracked/deleted results files
* Blocked on merging until @leszko's PR for redirection merges as I need to rebase on that branch to pick up some changes to e2e_test.go

test/e2e/e2e_test.go
```diff
@@ -224,10 +220,100 @@ func requireReplicatedStream(t *testing.T, c1 *catalystContainer, c2 *catalystCo
...
+		url     string
+		viewers int
+		timeout int
+		score   float64
+	}{
+		"hls": {url: fmt.Sprintf("http://localhost:%s/hls/stream+foo/index.m3u8", c2.http), viewers: 5, timeout: 30, score: 0.8},
+	}
+
+	// Test each protocol defined in the tests map above
+	for name, tc := range tests {
```
leszko, line 309: I see we have only one test in `tests`, do we plan to add more? If not, then do we need this loop at all? 🤔

test/e2e/e2e_test.go
```diff
@@ -224,10 +220,100 @@ func requireReplicatedStream(t *testing.T, c1 *catalystContainer, c2 *catalystCo
...
+		viewersPassed: vpass,
+		score:         vpass / vtotal,
+	}
+
+	return &r
+}
+
+// This test will spawn multiple viewers at node catalyst-two and attempt streaming.
+// Any hiccups in playback or failure to connect will be captured and reported via MistLoadTest.
+func requireProtocolLoad(t *testing.T, c1 *catalystContainer, c2 *catalystContainer) {
```
leszko, line 298: nit: Could you move this function up? It should be under `results` type definition. In general, it's better to "used" functions should be under "using" functions. To follow the [Stepdown rule](https://github.com/Geeksltd/Programming.Tips/blob/master/docs/methods/stepdown-rule.md).

test/e2e/e2e_test.go
```diff
@@ -224,10 +220,100 @@ func requireReplicatedStream(t *testing.T, c1 *catalystContainer, c2 *catalystCo
...
+		"-t", strconv.Itoa(timeout),
+		url)
+	out, err := cmdMistLoadTest.Output()
+	require.NoError(t, err)
+	if err != nil {
+		glog.Fatalf("Failed to start MistLoadTest: %s", err)
+	}
+	glog.Infof("MistLoadTest stdout: %s", out)
+
+	defer cmdMistLoadTest.Process.Kill()
```
leszko, line 255:
```suggestion
	defer cmdMistLoadTest.Process.Kill()
	out, err := cmdMistLoadTest.Output()
	require.NoError(t, err)
	if err != nil {
		glog.Fatalf("Failed to start MistLoadTest: %s", err)
	}
	glog.Infof("MistLoadTest stdout: %s", out)
```

leszko (approved):
Added some minor comments. Other than that, it looks good. Thanks for all the changes @emranemran 👍

test/e2e/e2e_test.go
emranemran: Yep, we'll be adding more streaming protocols to test in the future so decided to just have this framework from the get-go instead of refactoring again in the future.

test/e2e/e2e_test.go
emranemran: Fixed.

test/e2e/e2e_test.go
emranemran: Done

PR #61, opened 2022-07-27
Add playback redirection to the closest Mist node

### LINKED ISSUE #57: handle incoming stream redirection
We need to handle incoming requests for e.g. `https://playback.livepeer.studio/hls/PLAYBACK_ID/index.m3u8` and 302 redirect them to the address of a Catalyst server that has the stream in question. Previously, we proxied all of these requests behind the scenes, but that led to some suboptimal performance characteristics, especially around segment download latencies.

The workflow will be basically: 
1. Handle `GET https://playback.livepeer.studio/hls/PLAYBACK_ID/index.m3u8`
2. Turn around and ask MistUtilLoad for the best place to watch the pertinent stream, like so:
```
▶ curl 'http://localhost:8042/video+PLAYBACK_ID'
185.102.219.180
```
3. 302 Redirect the user to that URL `https://185.102.219.180/hls/PLAYBACK_ID/index.m3u8`
4. If no stream is present, 404

There will be some nuance in getting this right, I think... for example, we'll want to check for the `X-Forwarded-Proto: https` header and return an HTTPS url if present, that sort of thing.

@iameli TODO: Document why we're doing the flow this way

leszko:
fix #57


cmd/catalyst-node/catalyst-node.go
```diff
@@ -5,23 +5,29 @@ import (
 	"errors"
 	"flag"
 	"fmt"
+	"github.com/hashicorp/serf/cmd/serf/command/agent"
```
iameli, line 8: Hm, looks like our autoformatters disagree about where this line should be.

cmd/catalyst-node/catalyst-node.go
```diff
@@ -348,3 +359,66 @@ func main() {
 
 	os.Exit(exitCode)
 }
+
+func startCatalystWebServer(httpAddr string) {
+	http.Handle("/hls/", redirectHlsHandler())
+	glog.Infof("HTTP server listening on %s", httpAddr)
+	http.ListenAndServe(httpAddr, nil)
```
iameli, line 366: Should handle an error on the listener, yeah? Looks like this is a common pattern:
```suggestion
	glog.Fatalf(http.ListenAndServe(httpAddr, nil))
```

iameli: (But I didn't actually check that it doesn't crash the process with `FATAL ERROR: nil` so you should probably check that 😃)

cmd/catalyst-node/catalyst-node.go
```diff
@@ -348,3 +359,66 @@ func main() {
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		playbackID, isValid := parsePlaybackID(r.URL.Path)
+		if !isValid {
+			w.WriteHeader(http.StatusNotFound)
+			return
+		}
+
+		nodeAddr, err := getClosestNode(playbackID)
+		if err != nil {
+			glog.Error(err)
```
iameli, line 381:
```suggestion
			glog.Errorf("error finding origin server playbackID=%s error=%s", playbackID, err)
```

cmd/catalyst-node/catalyst-node.go
```diff
@@ -348,3 +359,66 @@ func main() {
...
+			return
+		}
+
+		redirectUrl := fmt.Sprintf("%s://%s/hls/%s/index.m3u8", protocol(r), nodeAddr, playbackID)
+		http.Redirect(w, r, redirectUrl, http.StatusFound)
+	})
+}
+
+func parsePlaybackID(path string) (string, bool) {
+	r := regexp.MustCompile("^/hls/([a-zA-Z0-9_+]+)/index.m3u8$")
```
iameli, line 392: Dashes are valid too:
```suggestion
	r := regexp.MustCompile("^/hls/([a-zA-Z0-9_\-+]+)/index.m3u8$")
```

cmd/catalyst-node/catalyst-node.go
```diff
@@ -348,3 +359,66 @@ func main() {
...
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return "", errors.New(fmt.Sprintf("GET request '%s' failed with http status code %d", url, resp.StatusCode))
+	}
+	body, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		return "", errors.New(fmt.Sprintf("GET request '%s' failed while reading response body", url))
+	}
+
```
iameli, line 422: If the stream isn't available — or (less likely) there are no servers with the resources to handle the playback, MistUtilLoad will return the string "FULL". Need to handle that error case, seems potentially appropriate for this function?

cmd/catalyst-node/catalyst-node_test.go
```diff
@@ -0,0 +1,80 @@
...
+	"fmt"
+	"github.com/stretchr/testify/require"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+)
+
+const (
+	closestNodeAddr = "someurl.com"
+	playbackID      = "abc123"
```
iameli, line 13: Just to handle a few valid cases, perhaps
```suggestion
	playbackID      = "abc_XYZ-123"
```

iameli (changes requested):
Looks solid, just a few tweaks. CI will pass if you merge/rebase off of master.

cmd/catalyst-node/catalyst-node.go
leszko: interesting. Updated back in 45710a545b44d9d636b16714b8055000286d6f92

cmd/catalyst-node/catalyst-node.go
leszko: Updated in 219d1f07fb8e7d35f8069c7af847ea8bb138de2e

leszko:
@iameli Addressed your comments. PTAL and review again.

iameli (approved):
LGTM! Merging to rebase some upstream PRs.

PR #63, opened 2022-07-28
Dockerfile: don't re-download on changes

iameli:


Dockerfile
```diff
@@ -8,9 +8,12 @@ ADD	go.mod go.sum	./
 
 RUN	go mod download
 
-ADD	.	.
+ADD	cmd/downloader/ cmd/downloader/
+ADD manifest.yaml manifest.yaml
+RUN go run cmd/downloader/main/downloader.go
```
hjpotter92, line 13:
```dockerfile
ADD	cmd/downloader/	cmd/downloader/
ADD	manifest.yaml Makefile ./
RUN	make download
```
the main script was moved to `cmd/downloader/downloader/downloader.go` so that i could read some functions from the manifest updater script as package

hjpotter92 (approved):

PR #79, opened 2022-08-02
Use `commit` info to populate inside manifest

hjpotter92:
This helps in negating docker cache; especially useful for bucket strategy builds.


iameli (approved):
LGTM!

PR #80, opened 2022-08-02
add latitude/longitude headers

iameli:


cmd/catalyst-node/catalyst-node.go
```diff
@@ -459,9 +462,20 @@ func protocol(r *http.Request) string {
...
+		return "", err
+	}
+	if lat != "" && lon != "" {
+		req.Header.Set("X-Latitude", lat)
+		req.Header.Set("X-Longitude", lon)
+	} else {
+		glog.Warningf("Incoming request missing X-Latitude/X-Longitude, response will not be geolocated")
+	}
+	client := &http.Client{}
+	resp, err := client.Do(req)
```
emranemran, line 478: Will this automatically use the lat/lon info or do we need to use the `--geo` flag to add weight to MistUtilLoad's scoring logic?

Issue #86, opened 2022-08-03
e2e_test: add failing test to detect issues with stream interruption


emranemran:
Particularly trying to detect failures shown in [\#65](https://github.com/DDVTECH/mistserver/issues/65)

PR #87, opened 2022-08-03
e2e_test: add new test to detect stream interruption failures

emranemran:
Bare minimum test case that reproduces this error: 
```
[2022-08-02 18:41:29] MistOutHLS:video+b0dcxvwml48mxt2s (490) INFO: Shared metadata not ready yet - no tracks valid (/tmp/drone-7pq2c6BO3H9JHJZs/drone/src/lib/dtsc.cpp:2151) 
``` 
Can be invoked by: 
``` 
go test -v test/e2e/* -run="Interrupt" --logtostderr
```

Fix \#86

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -111,7 +111,55 @@ func TestMultiNodeCatalyst(t *testing.T) {
 	requireStreamRedirection(t, c1, c2)
 }
 
-func TestMultiNodeCatalyst_BootstrapNode(t *testing.T) {
+func TestStreamInterruption(t *testing.T) {
+	if testing.Short() {
+		t.Skip("skipping testing in short mode")
+	}
+
+	// given
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	network := createNetwork(ctx, t)
+	defer network.Remove(ctx)
+
+	h1 := randomString("catalyst-")
+	h2 := randomString("catalyst-")
+
+	// when
+	c1 := startCatalyst(ctx, t, h1, network.name, defaultMistConfig())
+	defer c1.Terminate(ctx)
+	c2 := startCatalyst(ctx, t, h2, network.name, mistConfigConnectTo(h1))
+	defer c2.Terminate(ctx)
+
+	// then
+	requireMembersJoined(t, c1, c2)
+
+	// Begin ingesting and verify replication is working at catalyst-two node
+	p1 := startStream(t, c1)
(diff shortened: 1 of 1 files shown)


test/e2e/e2e_test.go
```diff
@@ -111,7 +111,55 @@ func TestMultiNodeCatalyst(t *testing.T) {
...
+	viewers := 5
+	timeout := 30
+	r := runProtocolLoadTest(t, "hls", fmt.Sprintf("http://localhost:%s/hls/stream+foo/index.m3u8", c2.http), viewers, timeout)
+	glog.Infof("Protocol under test: %v, viewers: %v, viewers-passed: %v, score: %v", r.protocol, r.viewers, r.viewersPassed, r.score)
+	if r.score == 0 {
+		glog.Infof("Failed %s test with score %v but ignoring failure", r.protocol, r.score)
+	} else {
+		t.Fatalf("Expected to fail but passed %s test with score: %v", r.protocol, r.score)
+	}
+	p2.Kill()
```
leszko, line 159: Move under `p2 := startStream(t, c1)` and call with `defer`

test/e2e/e2e_test.go
```diff
@@ -111,7 +111,55 @@ func TestMultiNodeCatalyst(t *testing.T) {
...
+	// Begin ingesting and verify replication is working at catalyst-two node
+	p1 := startStream(t, c1)
+	requireReplicatedStream(t, c2)
+
+	// Stop and then restart the ingest stream
+	p1.Kill()
+	p2 := startStream(t, c1)
+
+	// Start load test to simulate multiple viewers at catalyst-two node
+	// TODO: This test will succeed until issue #65 is resolved. The current
```
leszko, line 147: #65 is resolved already. Then, from what I understand, this test will fail. In that case, I suggest to not merge it right now.

leszko (changes requested):

test/e2e/e2e_test.go
emranemran: Sorry, meant issue 65 in mistserver repo: 
https://github.com/DDVTECH/mistserver/issues/65
I'll clarify in the comment. Also, the test is forced to pass even with a failure so that it will actually fail when the issue is fixed and we'll address the test correctly at that time.

test/e2e/e2e_test.go
emranemran: Will do

iameli (approved):

Issue #88, opened 2022-08-03
Makefile: "make mistserver" is currently broken


emranemran:
mistserver builds in catalyst is broken and reports the following when running `make mistserver`:
``` 
Builing release "Generic_64" for version "mist-v3.0-130-g3f81819" @ debug level 4
Building with SRT
Building without RIST support
SRT-LIB-IS-XXX="/home/parallels/catalyst/build/compiled/lib/libsrt.a"
-- Could NOT find Doxygen (missing: DOXYGEN_EXECUTABLE) 
Not cross compiling - building sourcery
-- Configuring done
-- Generating done
-- Build files have been written to: /home/parallels/mistserver
+ make -j2
make[1]: Entering directory '/home/parallels/catalyst/build/mistserver'
make[1]: *** No targets specified and no makefile found.  Stop.
make[1]: Leaving directory '/home/parallels/catalyst/build/mistserver'
make: *** [Makefile:60: mistserver] Error 2

```

iameli:
We figured it out - if you have an existing CMake build in the `mistserver` directory (e.g. you ran `cmake .` in `mistserver`) then it'll keep building in that location instead of the one within Catalyst. `git clean -ffdx` in `mistserver` fixes things.

PR #89, opened 2022-08-03
Makefile: fix 'make mistserver' builds

emranemran:
Fix \#88


hjpotter92 (approved):
lgtm

emranemran:
Not needed anymore for two reasons: 
* the issue was specific to my setup having an older CMakeCache.txt sitting inside mistserver repo (from doing builds there)
* we want to stick to separate repos instead of one giant monorepo with all relevant code bases.

PR #93, opened 2022-08-08
Add mistserver config file to 'clean' target

thomshutt:


iameli:
It's funny, the only reason the config file isn't in the directory itself is so `make clean` doesn't hit it 😆 

Maybe let's do a `make reset` that does `clean` and also wipes the config?

thomshutt:
Looks like you were one step ahead here! https://github.com/livepeer/catalyst/blob/main/Makefile#L155-L158

Issue #97, opened 2022-08-09
e2e_test: update stream interruption test to correctly fail if playback does not recover


emranemran:
In \#86, we added stream interruption test i.e. verify playback still recovers after ingest stream is killed and restarted. However, we set it to forcefully pass even with failures - this was temporary until we root-caused the issue. With \#65 now resolved, this stream interruption test can be correctly set to fail if playback is not able to recover correctly during an ingest stream interruption.

leszko:
@emranemran Are you still working on that?

PR #99, opened 2022-08-09
e2e_test: set stream-interruption test to fail correctly

emranemran:
In #86, we added stream interruption test i.e. verify playback still recovers
after ingest stream is killed and restarted. However, we set it to forcefully
pass even with failures - this was temporary until we root-caused the issue.
With #65 now resolved, this stream interruption test can be correctly set to
fail if playback is not able to recover correctly during an ingest stream
interruption.

Fix \#97

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -46,6 +46,7 @@ box:
         darwin-arm64: livepeer-mistserver-darwin-arm64.tar.gz
         linux-amd64: livepeer-mistserver-linux-amd64.tar.gz
         linux-arm64: livepeer-mistserver-linux-arm64.tar.gz
+        #TODO: automate mist builds
     - name: www
       strategy:
         download: github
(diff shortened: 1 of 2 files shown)


manifest.yaml
```diff
@@ -46,6 +46,7 @@ box:
         darwin-arm64: livepeer-mistserver-darwin-arm64.tar.gz
         linux-amd64: livepeer-mistserver-linux-amd64.tar.gz
         linux-arm64: livepeer-mistserver-linux-arm64.tar.gz
+        #TODO: automate mist builds
```
leszko, line 49: What is this comment? I think it's not clear enough, so I'd rather recommend removing it (and create a GH Issue instead).

leszko (approved):
Added one comment. Other than that, LGTM 👍

PR #103, opened 2022-08-10
docker: Expose port 7979 from catalyst-api

victorges:


hjpotter92:
is the `EXPOSE` really needed? wouldn't port-binding supersede any values specified in image

victorges:
> @hjpotter92 is the EXPOSE really needed? wouldn't port-binding supersede any values specified in image

So this was one attempt of fixing the `connection refused` I was getting, but it was likely not (only) it but that server listening on `localhost`. What do you mean with "port-binding" though? I know that we do not declare any ports in the k8s pod, if that's what you mean.

hjpotter92:
The `EXPOSE` clause works only as a documentation, and is not of any consequences when container gets deployed (in k8s/compose/swarm/anywhere).

I believe the issue with catalyst-api would have been only the server binding to `localhost` instead of `0.0.0.0`.

We can close this PR as a no-op? unless caalyst-api still needs to be exposed publicly (instead of internally)

---

port-binding: I meant to say binding/publishing port from host to container, in k8s this is achieved by the `Service` manifests' port mapping :)

victorges:
Yeah can close this!

PR #115, opened 2022-08-17
S3 output support: add catalyst-uploader binary

cyberj0g:
Added `bin/` dir to PATH for `dev` target, because otherwise Mist can't find `livepeer-catalyst-uploader` binary. Should we update binary discovery in Mist?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -118,7 +118,7 @@ manifest:
 
 .PHONY: dev
 dev:
-	if [ $$(uname) == "Darwin" ]; then \
+	if [ $$(uname) = "Darwin" ]; then \
 		if [ ! -d "/Volumes/RAMDisk/" ]; then \
 			disk=$$(hdiutil attach -nomount ram://4194304) \
 			&& sleep 3 \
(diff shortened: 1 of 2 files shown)


leszko (approved):

manifest.yaml
```diff
@@ -38,7 +45,7 @@ box:
       strategy:
         download: bucket
         project: mistserver
-        commit: 3b2cb3abaf2c107649133608cb07806e4eecdba2
+        commit: 5f695e47bfa8f42dce7bd25c6d78852cc4f300e9
```
hjpotter92, line 48: this seems to be a rebase issue? try rebasing from current main branch

Makefile
```diff
@@ -128,6 +128,7 @@ dev:
 		&& rm -rf /Volumes/RAMDisk/mist \
 		&& export TMP=/Volumes/RAMDisk; \
 	fi \
+	&& export PATH=$$PATH:$$(pwd)/bin \
```
hjpotter92, line 131: i don't know if the `PATH` gets exported to the shell calling make commands or is it only for make's context. I'd avoid modifying system's `PATH` value here

thomshutt (approved):

PR #122, opened 2022-08-25
catalyst-node: Add `redirectPrefixes` support when querying mist

### LINKED ISSUE #95: catalyst-node: handle `video` and `videorec` prefixes
Currently stream in Livepeer Studio can have two prefixes, corresponding to two Mist stream configurations: `video` for normal streams, and `videorec` for streams with recording enabled.

So our playback URLs look like this:
```
https://catalyst-node.example.com/hls/video+PLAYBACK_ID/index.m3u8
https://catalyst-node.example.com/hls/videorec+PLAYBACK_ID/index.m3u8
```

But in the dashboard, our playback URLs don't have any prefixes, and look like this:
```
https://catalyst-node.example.com/hls/PLAYBACK_ID/index.m3u8
```

The legacy system uses etcd routes and middleware in Traefik to handle this sort of thing, but we're getting rid of that. Other Mist configurations use the `defaultStream` parameter to automatically rewrite `PLAYBACK_ID` to e.g. `stream+PLAYBACK_ID`, but that won't work for us because we have two stream configurations, `video` and `videorec`.

I'm leaning toward handling this on the catalyst-node side, so it can issue a redirect from `https://global-playback-url.example.com/hls/PLAYBACK_ID/index.m3u8` to `https://catalyst-node.example.com/hls/PLAYBACK_ID/index.m3u8`.

**New command line option**: `-redirectPrefixes=video,videorec`. Then when we get an incoming playback request for `PLAYBACK_ID`, we'll query the Mist load balancer concurrently for `video+PLAYBACK_ID` and `videorec+PLAYBACK_ID`. (If an incoming request comes in for `video+playbackId` we can just forward it as-is.) Then we can issue a redirect to whichever one has a stream, if any.

hjpotter92:
Fixes #95 and livepeer/livepeer-infra#894

### DIFF
--- a/cmd/catalyst-node/catalyst-node.go
+++ b/cmd/catalyst-node/catalyst-node.go
@@ -305,7 +305,7 @@ func main() {
 	}
 
 	cliFlags.RedirectPrefixes = strings.Split(*prefixes, ",")
-	glog.Infof("found redirectPrefixes=%v", cliFlags.RedirectPrefixes)
+	glog.V(4).Infof("found redirectPrefixes=%v", cliFlags.RedirectPrefixes)
 
 	parseSerfConfig(&serfConfig, retryJoin, serfTags)
 
--- a/cmd/catalyst-node/catalyst-node.go
+++ b/cmd/catalyst-node/catalyst-node.go
@@ -305,6 +305,7 @@ func main() {
 	}
 
 	cliFlags.RedirectPrefixes = strings.Split(*prefixes, ",")
+	glog.Infof("found redirectPrefixes=%v", cliFlags.RedirectPrefixes)
 
 	parseSerfConfig(&serfConfig, retryJoin, serfTags)
 
@@ -443,7 +444,7 @@ func redirectHlsHandler(redirectPrefixes []string) http.Handler {
 		lat := r.Header.Get("X-Latitude")
 		lon := r.Header.Get("X-Longitude")
 
-		var nodeAddr string
+		var nodeAddr, validPrefix string
 		var err error
 
 		for _, prefix := range redirectPrefixes {
@@ -453,6 +454,7 @@ func redirectHlsHandler(redirectPrefixes []string) http.Handler {
 				continue
 			}
 			if nodeAddr != "" {
+				validPrefix = prefix + "+"
 				err = nil
 				break
 			}
@@ -464,7 +466,7 @@ func redirectHlsHandler(redirectPrefixes []string) http.Handler {
 			return
 		}
 
-		rURL := fmt.Sprintf("%s://%s/hls/%s/index.m3u8", protocol(r), nodeAddr, playbackID)
+		rURL := fmt.Sprintf("%s://%s/hls/%s%s/index.m3u8", protocol(r), nodeAddr, validPrefix, playbackID)
 		http.Redirect(w, r, rURL, http.StatusFound)
 	})
 }
--- a/cmd/catalyst-node/catalyst-node.go
+++ b/cmd/catalyst-node/catalyst-node.go
@@ -308,7 +308,7 @@ func main() {
 
 	parseSerfConfig(&serfConfig, retryJoin, serfTags)
 
-	go startCatalystWebServer(cliFlags.HTTPAddress)
+	go startCatalystWebServer(cliFlags.HTTPAddress, cliFlags.RedirectPrefixes)
 
 	config.serfTags = serfConfig.Tags
 
@@ -325,7 +325,7 @@ func main() {
 		// eli note: i put this in a loop in case client boots before server.
 		// doesn't seem to happen in practice.
 		for {
-			err := runClient(config)
+			err := runClient(*config)
 			if err != nil {
 				glog.Errorf("Error starting client: %v", err)
 			}
--- a/cmd/catalyst-node/catalyst-node.go
+++ b/cmd/catalyst-node/catalyst-node.go
@@ -13,6 +13,7 @@ import (
 	"regexp"
 	"runtime"
 	"strings"
+	"sync"
 	"time"
 
 	serfclient "github.com/hashicorp/serf/client"
@@ -446,19 +447,24 @@ func redirectHlsHandler(redirectPrefixes []string) http.Handler {
 
 		var nodeAddr, validPrefix string
 		var err error
+		var waitGroup sync.WaitGroup
 
 		for _, prefix := range redirectPrefixes {
-			nodeAddr, err = getClosestNode(playbackID, lat, lon, prefix)
-			if err != nil {
-				glog.Errorf("error finding origin server playbackID=%s prefix=%s error=%s", playbackID, prefix, err)
-				continue
-			}
-			if nodeAddr != "" {
-				validPrefix = prefix + "+"
-				err = nil
-				break
-			}
+			waitGroup.Add(1)
+			go func(prefix string) {
+				addr, err := getClosestNode(playbackID, lat, lon, prefix)
+				if err != nil {
+					glog.V(8).Infof("error finding origin server playbackID=%s prefix=%s error=%s", playbackID, prefix, err)
+				}
+				if addr != "" {
+					nodeAddr = addr
+					validPrefix = prefix + "+"
+					err = nil
+				}
+				waitGroup.Done()
+			}(prefix)
 		}
+		waitGroup.Wait()
 
 		if nodeAddr == "" || err != nil {
 			glog.Errorf("error finding origin server playbackID=%s error=%s", playbackID, err)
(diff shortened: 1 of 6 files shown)


cmd/catalyst-node/catalyst-node.go
```diff
@@ -452,7 +461,8 @@ func parsePlaybackID(path string) (string, bool) {
 	if len(m) < 2 {
 		return "", false
 	}
-	return m[1], true
+	slice := strings.Split(m[1], "+")
```
leszko, line 464: I don't understand this part. Why do we split by `+`? Maybe add a comment with the explanation.

cmd/catalyst-node/catalyst-node.go
```diff
@@ -462,35 +472,51 @@ func protocol(r *http.Request) string {
...
-	} else {
-		glog.Warningf("Incoming request missing X-Latitude/X-Longitude, response will not be geolocated")
-	}
-	client := &http.Client{}
-	resp, err := client.Do(req)
-	if err != nil {
-		return "", err
-	}
-	defer resp.Body.Close()
+func queryMistForClosestNode(playbackID, lat, lon string, redirectPrefixes []string) (string, error) {
```
leszko, line 475: I think you could improve this part by doing the following:
1. Change this function to `queryMistForClosestNode(playbackID, lat, lon string, redirectPrefix string) (string, error)`
2. Move the for loop outside this function `for _, prefix := range redirectPrefixes`
3. Then, you could add the test for the logic you added.

What do you think?

leszko (approved):
@hjpotter92 Thanks for the PR. Added 2 comments. Other than that LGTM.

I haven't tested it with Mist, so if you haven't yet, I think you should give it a shot before merging.

Thanks!

cmd/catalyst-node/catalyst-node.go
hjpotter92: Added a comment explaining this.

cmd/catalyst-node/catalyst-node.go
hjpotter92: yeah, that is what i should have done in the first place. wrote most of the code while in zombie mode yesterday. updated now :+1:

cmd/catalyst-node/catalyst-node.go
```diff
@@ -434,25 +444,47 @@ func redirectHlsHandler() http.Handler {
...
 
-		nodeAddr, err := getClosestNode(playbackID, lat, lon)
-		if err != nil {
+		var nodeAddr, validPrefix string
+		var err error
+
+		for _, prefix := range redirectPrefixes {
+			nodeAddr, err = getClosestNode(playbackID, lat, lon, prefix)
+			if err != nil {
+				glog.Errorf("error finding origin server playbackID=%s prefix=%s error=%s", playbackID, prefix, err)
```
leszko, line 453: I think we should not log error if we got just one error for one prefix. IFAIU the situation in which Mist does not have `video+<playback_id>` but has `videorec+<playback_id` is completely normal and should not result in an error.

cmd/catalyst-node/catalyst-node.go
```diff
@@ -434,25 +444,47 @@ func redirectHlsHandler() http.Handler {
...
-		rURL := fmt.Sprintf("%s://%s/hls/%s/index.m3u8", protocol(r), nodeAddr, playbackID)
+		rURL := fmt.Sprintf("%s://%s/hls/%s%s/index.m3u8", protocol(r), nodeAddr, validPrefix, playbackID)
+		glog.V(6).Infof("generated redirect url=%s", rURL)
 		http.Redirect(w, r, rURL, http.StatusFound)
 	})
 }
 
 func parsePlaybackID(path string) (string, bool) {
-	r := regexp.MustCompile("^/hls/([a-zA-Z0-9_\\-+]+)/index.m3u8$")
+	r := regexp.MustCompile("^/hls/([\\w+-]+)/index.m3u8$")
```
leszko, line 476: Good change! 👏 

One thing I wonder is if it shouldn't be `\\w\\-+`? Don't we need to include `\`?

cmd/catalyst-node/catalyst-node_test.go
```diff
@@ -14,21 +14,23 @@ const (
 	playbackID      = "abc_XYZ-123"
 )
 
+var prefixes = [...]string{"video"}
```
leszko, line 17: I think you should add a test case for multiple prefixes `[...]string{"video", "videorec"}` in which Mist don't find anything for `video`, but it finds for `videorec`.

cmd/catalyst-node/catalyst-node.go
hjpotter92: makes sense. will update this.

cmd/catalyst-node/catalyst-node_test.go
hjpotter92: since we override the `getClosestNode` method, it always returns `video` as the prefix from the loop

cmd/catalyst-node/catalyst-node.go
hjpotter92: We aren't including `\`; it was just to escape the `-` character in original pattern. Otherwise it'd be interpreted as character set from `_` to `+`; moving `-` to beginning or end of character set doesnt need to escape it.

iameli: Are underscores still included here? They should be.

cmd/catalyst-node/catalyst-node.go
```diff
@@ -235,24 +244,27 @@ func execBalancer(balancerArgs []string) error {
 }
 
 func main() {
+	var cliFlags = &catalystNodeCliFlags{}
```
iameli, line 247: Yayyyy I was just thinking it was time for us to roll to this

cmd/catalyst-node/catalyst-node.go
```diff
@@ -284,28 +296,26 @@ func main() {
...
+	if cliFlags.Version {
 		fmt.Println("catalyst-node version: " + Version)
 		fmt.Printf("golang runtime version: %s %s\n", runtime.Compiler, runtime.Version())
 		fmt.Printf("architecture: %s\n", runtime.GOARCH)
 		fmt.Printf("operating system: %s\n", runtime.GOOS)
 		return
 	}
 
+	cliFlags.RedirectPrefixes = strings.Split(*prefixes, ",")
+	glog.Infof("found redirectPrefixes=%v", cliFlags.RedirectPrefixes)
```
iameli, line 308: Maybe a bit less verbose?
```suggestion
	glog.V(6).Infof("found redirectPrefixes=%v", cliFlags.RedirectPrefixes)
```

cmd/catalyst-node/catalyst-node.go
```diff
@@ -434,25 +444,47 @@ func redirectHlsHandler() http.Handler {
 		lat := r.Header.Get("X-Latitude")
 		lon := r.Header.Get("X-Longitude")
 
-		nodeAddr, err := getClosestNode(playbackID, lat, lon)
-		if err != nil {
+		var nodeAddr, validPrefix string
+		var err error
+
+		for _, prefix := range redirectPrefixes {
```
iameli, line 450: I think we should probably parallelize this with a waitGroup — otherwise the `videorec` has to wait for a round-trip for `video` every single time

cmd/catalyst-node/catalyst-node.go
```diff
@@ -462,8 +494,8 @@ func protocol(r *http.Request) string {
 	return "http"
 }
 
-func queryMistForClosestNode(playbackID, lat, lon string) (string, error) {
-	url := fmt.Sprintf("http://localhost:%s/%s", mistUtilLoadPort, playbackID)
+func queryMistForClosestNode(playbackID, lat, lon, prefix string) (string, error) {
+	url := fmt.Sprintf("http://localhost:%d/%s+%s", mistUtilLoadPort, prefix, playbackID)
```
iameli, line 498: Hmmmmmmmmmmm I'm trying to decide if we want to support non-prefixed playback. I think it's fine that we don't in this PR but I can imagine someone wanting to do something with that in the future

iameli (changes requested):
I do think we should parallelize the prefix searches but otherwise LGTM

cmd/catalyst-node/catalyst-node.go
hjpotter92: yes, `\w` comes from perl; consisting of `[a-zA-Z0-9_]`. referring from docs at: https://golang.org/s/re2syntax

cmd/catalyst-node/catalyst-node.go
hjpotter92: verbosity logs only support `Info` messages. updated this in next commit.

cmd/catalyst-node/catalyst-node.go
hjpotter92: done!

![i][i]

[i]: https://media2.giphy.com/media/AvMJCeu1EMmhG/giphy.gif

iameli (approved):
Looks great!

PR #136, opened 2022-09-06
Add support for segments in the redirection mechanism

leszko:
Add support for the segment in the playback path. For example, `https://playback.livepeer.fun/hls/5797b1te5y6au14l/0_1/index.m3u8`

Before we only supported a path without segments, for example, `https://playback.livepeer.fun/hls/5797b1te5y6au14l/index.m3u8`.

That broke our loadtester and record-tester.

fix https://github.com/livepeer/livepeer-infra/issues/927

### DIFF
--- a/cmd/catalyst-node/catalyst-node.go
+++ b/cmd/catalyst-node/catalyst-node.go
@@ -435,7 +435,7 @@ func redirectHlsHandler(redirectPrefixes []string) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("Access-Control-Allow-Origin", "*")
 		w.Header().Set("Access-Control-Allow-Headers", "*")
-		playbackID, isValid := parsePlaybackID(r.URL.Path)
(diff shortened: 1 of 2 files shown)


iameli (changes requested):
I don't think this will have the effect that we want. This will ask the Mist load balancer for a server for every single segment request that comes in. That means individual segments will get routed to different servers as network conditions change. Each one of those will trigger its own session authentication and whatnot - messy and high-latency.

Additionally, the Mist load balancer is (partially) stateful, and running each segment request through here will cause it to think that each segment request is kicking off a new user playback of the stream, which will add up and quickly exhaust the available bandwidth.

I think this needs to be handled on the stream-tester side by detecting that a 301/302 redirect has taken place and re-targeting the playback request on the new server. Sorry for the confusion!

iameli (approved):
Changing my ❌ to a ✅ per discussion in Discord. We still need to do the stream-tester thing, but this behavior is useful for playback recovery upon server failure.

ya7ya (approved):
LGTM

PR #140, opened 2022-09-08
catalyst-node: add support for /json_prefix+stream.js urls, 404 handling

iameli:
I'm kind of unhappy with the messiness of this; in particular returning both a URL and an error from the `queryMistForClosestNode()` function is kind of smelly. I at least tried to thoroughly-comment all the confusing cases of "we didn't find the stream but return the server anyway" kind of thing.

### DIFF
--- a/cmd/catalyst-node/catalyst-node.go
+++ b/cmd/catalyst-node/catalyst-node.go
@@ -431,6 +431,46 @@ func startCatalystWebServer(httpAddr string, redirectPrefixes []string) {
 
 var getClosestNode = queryMistForClosestNode
 
+// return the best node available for a given stream. will return any node if nobody has the stream.
+func getBestNode(redirectPrefixes []string, playbackID, lat, lon string) (string, string, error) {
+	var nodeAddr, fullPlaybackID, fallbackAddr string
+	var err error
+	var waitGroup sync.WaitGroup
+
+	for _, prefix := range redirectPrefixes {
+		waitGroup.Add(1)
+		go func(prefix string) {
+			addr, e := getClosestNode(playbackID, lat, lon, prefix)
+			if e != nil {
+				err := e
+				glog.V(8).Infof("error finding origin server playbackID=%s prefix=%s error=%s", playbackID, prefix, err)
+				// If we didn't find a stream but we did find a server, keep that so we can use it to handle a 404
+				if addr != "" {
+					fallbackAddr = addr
+				}
+			} else {
+				nodeAddr = addr
+				fullPlaybackID = prefix + "+" + playbackID
+			}
+			waitGroup.Done()
+		}(prefix)
+	}
+	waitGroup.Wait()
+
+	// good path: we found the stream and a good node to play it back, yay!
+	if nodeAddr != "" {
+		return nodeAddr, fullPlaybackID, nil
+	}
+
+	// bad path: nobody has the stream, but we did find a server which can handle the 404 for us.
+	if fallbackAddr != "" {
+		return fallbackAddr, redirectPrefixes[0] + "+" + playbackID, nil
+	}
+
+	// ugly path: we couldn't find ANY servers. yikes.
+	return "", "", err
+}
+
 func redirectHandler(redirectPrefixes []string) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		w.Header().Set("Access-Control-Allow-Origin", "*")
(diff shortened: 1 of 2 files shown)


cmd/catalyst-node/catalyst-node.go
```diff
@@ -424,18 +424,58 @@ func writeSerfConfig(config *agent.Config) (string, error) {
...
+	http.Handle("/", redirectHandler(redirectPrefixes))
 	glog.Infof("HTTP server listening on %s", httpAddr)
 	glog.Fatal(http.ListenAndServe(httpAddr, nil))
 }
 
 var getClosestNode = queryMistForClosestNode
 
-func redirectHlsHandler(redirectPrefixes []string) http.Handler {
+// return the best node available for a given stream. will return any node if nobody has the stream.
+func getBestNode(redirectPrefixes []string, playbackID, lat, lon string) (string, string, error) {
```
leszko, line 435: Nit: I suggest to move this function under `redirectHandler()`. It's simpler to read when you follow the [Stepdown Rule](https://dzone.com/articles/the-stepdown-rule).

cmd/catalyst-node/catalyst-node.go
```diff
@@ -424,18 +424,58 @@ func writeSerfConfig(config *agent.Config) (string, error) {
...
+	var nodeAddr, fullPlaybackID, fallbackAddr string
+	var err error
+	var waitGroup sync.WaitGroup
+
+	for _, prefix := range redirectPrefixes {
+		waitGroup.Add(1)
+		go func(prefix string) {
+			addr, e := getClosestNode(playbackID, lat, lon, prefix)
+			if e != nil {
+				err := e
```
leszko, line 445: This line is wrong. I guess what you wanted to have is `err = e`. Otherwise, you just create a new variable, which is not used anywhere.

On the other hand if you use `err = e`, then you need to think how to sync it with the other goroutine.

cmd/catalyst-node/catalyst-node.go
```diff
@@ -501,12 +549,19 @@ func protocol(r *http.Request) string {
 
 func queryMistForClosestNode(playbackID, lat, lon, prefix string) (string, error) {
 	// First, check to see if any server has this stream
-	_, err := queryMistForClosestNodeSource(playbackID, lat, lon, prefix, true)
-	if err != nil {
-		return "", err
-	}
-	// If so, fetch the best server for playback
-	return queryMistForClosestNodeSource(playbackID, lat, lon, prefix, false)
+	_, err1 := queryMistForClosestNodeSource(playbackID, lat, lon, prefix, true)
```
leszko, line 552: This logic seems fishy. Shouldn't we just do like this?
1. Check the best playback server (`source = false`)
2. Only if there is no best playback server, then query for any (`source = true`)

Now, we always query for both which seems kind-of weird. If it needs to be that way, then I suggest to add a good comment which explains why it is so.

cmd/catalyst-node/catalyst-node.go
```diff
@@ -489,7 +510,34 @@ func parsePlaybackID(path string) (string, string, bool) {
...
+	return slice[len(slice)-1], pathTmpl, true
+}
+
+func parsePlaybackIDJS(path string) (string, string, bool) {
+	r := regexp.MustCompile(`^/json_([\w+-]+).js$`)
+	m := r.FindStringSubmatch(path)
+	if len(m) < 2 {
+		return "", "", false
+	}
+	// Incoming requests might come with some prefix attached to the
```
leszko, line 523: I think this comment is not correct. Looks like copy-paste.

cmd/catalyst-node/catalyst-node.go
```diff
@@ -424,18 +424,58 @@ func writeSerfConfig(config *agent.Config) (string, error) {
...
+
+	for _, prefix := range redirectPrefixes {
+		waitGroup.Add(1)
+		go func(prefix string) {
+			addr, e := getClosestNode(playbackID, lat, lon, prefix)
+			if e != nil {
+				err := e
+				glog.V(8).Infof("error finding origin server playbackID=%s prefix=%s error=%s", playbackID, prefix, err)
+				// If we didn't find a stream but we did find a server, keep that so we can use it to handle a 404
+				if addr != "" {
```
leszko, line 448: I added a similar comment somewhere, but I think this logic looks fishy. I think here is the point where we should query for the fallback node (meaning for any node). 

However we query for ANY node in `getCloseNode` which seems very confusing.

leszko (changes requested):

cmd/catalyst-node/catalyst-node.go
iameli: * The `source = false` call always returns a server even if the stream isn't live anywhere. We need that server to be the 404 handler.
* The `source=true` call only returns successfully if the stream is live somewhere. We need this information to know whether we should be redirecting to `video` or `videorec`.

Don't see how we can avoid doing both, but I can add a comment to that effect

cmd/catalyst-node/catalyst-node.go
iameli: Good catch, wasn't really sure how to handle this case. I guess I was thinking that I didn't really care which error that I returned from the outer error handler; presumably the requests failed for the same reason and we could just return any of them. So maybe I don't care about the sync. But it's not very elegant for sure.

leszko: I think that in Go, you MUST care about the sync. Because the assignment operation itself is not atomic. So, in the worse case scenario, you may end up here with some panic / seg fault.

cmd/catalyst-node/catalyst-node.go
iameli: Right - we do want the plus-splitting logic in both cases but the comment isn't quite right. I'll try and clean up both.

cmd/catalyst-node/catalyst-node.go
iameli: Oooh, I didn't know that. That'd be a problem even before my change then, yeah? Both goroutines might concurrently modify `nodeAddr`... do you have an idea for a better pattern here? I want something like `Promise.all` in JavaScript where I can get a list of the errors/results afterward, I think...

cmd/catalyst-node/catalyst-node.go
iameli: Hmm, it might be just be the case that we have the wrong prefix though; we can't be sure there's not a solid match for the stream elsewhere. I agree this logic is confusing/feels messy. I think the right way for us to do this wouldn't be to query Mist at all, and instead use our own list of servers to return a fallback when appropriate. But we don't yet have any geolocation logic or anything outside of MistUtilLoad, so it's tricky.

cmd/catalyst-node/catalyst-node.go
leszko: Yeah, I think you're right. That the problem was there even before you changes. You can just use a mutex to solve it.

Other options are channels or some library to have atomic string, but I think it's an overkill for your simple scenario.

iameli: Added a mutex

PR #106, opened 2022-08-15
Add vmagent to manifest, better scrape_configs

ya7ya:
this is required to export logs to the remote victoria-metrics clusters we have

PR #141, opened 2022-09-09
Add auto push of live streams, add MinIO S3 dev dependency

cyberj0g:
**Changes:**
* Added Mist dev configuration to push new live streams (with names starting with `video+`) to S3.
* Added MinIO docker container to make it easier to work with S3 storage locally

Related [issue](https://github.com/livepeer/go-livepeer/issues/2506). 

**In progress:**
- test with Catalyst Studio, getting errors trying to run it locally

**Need help with:**
- updating `autopush` section of Mist config for production, including adding S3 secrets

### DIFF
--- a/config/minio.sh
+++ b/config/minio.sh
@@ -4,7 +4,13 @@ minio server /data --console-address ":9001"&
 sleep 5
 # download management utility
 if [[ ! -e "/data/mc" ]]; then
-  curl https://dl.min.io/client/mc/release/linux-amd64/mc --create-dirs -o /data/mc
+  case $(uname -m) in
+    x86_64) ARCH="amd64" ;;
+    arm64)  ARCH="arm64" ;;
+    *) ARCH="$(uname -m)" ;;
+  esac
+  OSARCH=$(uname|tr [:upper:] [:lower:])-$ARCH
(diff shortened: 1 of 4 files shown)


iameli (changes requested):
Hmmm — I think for Minio I'd actually prefer that we package it with the Catalyst build itself, like we do with Victoria Metrics and vmagent — see @ya7ya's PR here. Seems like a capability that would be good to ship alongside Catalyst. Postgres and Rabbit are both in a different category; they're Studio-only dependencies that have a lot of other moving parts (such as RabbitMQ's Erlang development environment.) https://github.com/livepeer/catalyst/pull/106

config/minio.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+# start minio server
+minio server /data --console-address ":9001"&
+sleep 5
+# download management utility
+if [[ ! -e "/data/mc" ]]; then
+  curl https://dl.min.io/client/mc/release/linux-amd64/mc --create-dirs -o /data/mc
```
hjpotter92, line 7: should this be using `uname -m` instead of `amd64`? especially for arm64 macos machines?

cyberj0g: good catch, fixed according to minio's naming convention

Issue #110, opened 2022-08-15
build process packages up old versions of MistServer somehow


iameli:
Check out [this build](https://github.com/livepeer/catalyst/runs/7843530191?check_suite_focus=true). It's for commit cc9d1881159057f782222e725d03a8f3dd047634, right? [Here's manifest.yaml for that commit](https://github.com/livepeer/catalyst/blob/cc9d1881159057f782222e725d03a8f3dd047634/manifest.yaml#L41), showing MistServer version DDVTECH/mistserver@2c3331192168d1c68863aef7320b60cd920d8baa.

However, the build itself downloads DDVTECH/mistserver@52e1cb9c35555ca67ab215d3f80919b344fb0c32. This shows up in the logs when it does verification:
```
#24 120.5 I0815 18:17:07.185310     474 downloader.go:67] Doing SHA checksum verification for service=mistserver
#24 120.7 I0815 18:17:07.375632     474 shasum.go:13] Verifying shasum with file=52e1cb9c35555ca67ab215d3f80919b344fb0c32_checksums.txt
```

I'm guessing the offending error is this, occurring right at the `git clone` step?

```
fatal: Not a valid object name cc9d1881159057f782222e725d03a8f3dd047634^{commit}
```

hjpotter92:
My initial thought was that it might be timing issue, where the mist builds had not yet succeeded and catalyst build was triggered. However, we push the updated manifest only after the binaries have been successfully compressed and uploaded to bucket; which in-turn created the auto-commit.

The other alternative "could" be that github action fetched the old copy of manifest (is there cloudflare caching on `build.livepeer.live` domain?)

The 3rd possibility i can think of is that the `staging` branch was merged pretty quick which led to the builder (which uses github URL for context) wasn't able to fetch information about it? This, however, shouldn't affect the `make build` pipeline at all.

Creating a PR with more logging context. I'll be observing the builds from auto-commits and their merges closely.

hjpotter92:
This seems to have been a caching issue. A fix was deployed to cloudflare yesterday; and the worker will be moved to wrangler with this open PR: livepeer/livepeer-worker#6

leszko:
@hjpotter92 If the related PR is merged, can we close this issue?

PR #143, opened 2022-09-12
bucket.go: Use commit hash provided in manifest over current value

hjpotter92:
This would allow for finer version control on released binaries. Currently, the downloads are always done for most 'recent' commit available for a branch in `bucket` strategy. This led to and old issue #110 

With the change here, the commit SHA (if present in `manifest.yaml`) would be the default value to download.


leszko (approved):

iameli (approved):
Yup, this seems correct.

PR #150, opened 2022-09-13
catalyst-node: changes for production release

iameli:
A bunch of changes as we close in on the production deployment:

* Add `-advertise`, allowing catalyst-node to advertise an IP address that it's not actually binding to (needed for Kubernetes `hostPort` support)
* Add `-mist-load-balancer-template`, allowing us to specify alternate ports for Mist to query nodes. (needed for running Canary stuff on `:20443`)
* https://github.com/livepeer/catalyst/pull/150/commits/32be8fdb18513e27da12f8782251b2ae810c13a1, which checks `media=true` nodes for the load balancer (instead of matching the local node's tags)
* Added the ability for nodes to specify `http`, `https`, and `dtsc` tags, specifying non-default ports and hosts for connections. We're using this for the new rollout system; a new set of Catalyst pods can roll out on the same IPs with DTSC listening on (for example) 4201 instead of 4200 and traffic will get routed to them as appropriate.
  * To implement this, I've added a new internal HTTP server on port 8091 for handling internal Mist stream triggers and that sort of thing, similar to go-livepeer's 7935 server.

This is a little ugly right now - mostly the global `serfClient` variable. We're probably due for a refactor to give everything in here proper context, split into multiple files, etc. If anyone can think of a more elegant way to handle the `serfClient` situation sort of a full refactor, that'd be good.

### DIFF
--- a/cmd/catalyst-node/catalyst-node.go
+++ b/cmd/catalyst-node/catalyst-node.go
@@ -544,14 +544,7 @@ func resolveNodeURL(streamURL string) (string, error) {
 	nodeName := u.Host
 	protocol := u.Scheme
 
-	members, err := serfClient.MembersFiltered(map[string]string{}, "alive", nodeName)
-	if err != nil {
-		return "", err
-	}
-	if len(members) < 1 {
-		return "", fmt.Errorf("serf node not found name=%s", nodeName)
-	}
-	member := members[0]
+	member, err := getSerfMember(nodeName)
 	addr, has := member.Tags[protocol]
 	if !has {
 		glog.V(7).Infof("no tag found, not tag resolving protocol=%s nodeName=%s", protocol, nodeName)
@@ -565,6 +558,22 @@ func resolveNodeURL(streamURL string) (string, error) {
 	u2.RawQuery = u.RawQuery
 	return u2.String(), nil
 }
+
+func querySerfForMember(name string) (*serfclient.Member, error) {
+	members, err := serfClient.MembersFiltered(map[string]string{}, "alive", name)
+	if err != nil {
+		return nil, err
+	}
+	if len(members) < 1 {
+		return nil, fmt.Errorf("could not find serf member name=%s", name)
+	}
+	if len(members) > 1 {
+		glog.Errorf("WARNING: found multiple serf members with the same name! name=%s count=%d", name, len(members))
+	}
+	return &members[0], nil
+}
+
+var getSerfMember = querySerfForMember
 
 // return the best node available for a given stream. will return any node if nobody has the stream.
 func getBestNode(redirectPrefixes []string, playbackID, lat, lon string) (string, string, error) {
--- a/cmd/catalyst-node/catalyst-node.go
+++ b/cmd/catalyst-node/catalyst-node.go
@@ -507,7 +507,7 @@ func streamSourceHandler() http.Handler {
 		b, err := io.ReadAll(r.Body)
 		if err != nil {
 			glog.Errorf("error handling STREAM_SOURCE body=%s", err)
-			w.WriteHeader(http.StatusInternalServerError)
+			w.Write([]byte("push://"))
 			return
 		}
 		streamName := string(b)
@@ -515,7 +515,7 @@ func streamSourceHandler() http.Handler {
 		source, err := getStreamSource(streamName)
 		if err != nil {
 			glog.Errorf("error finding STREAM_SOURCE: %s", err)
-			w.WriteHeader(http.StatusInternalServerError)
+			w.Write([]byte("push://"))
 			return
 		}
 		glog.V(7).Infof("replying to Mist STREAM_SOURCE request=%s response=%s", streamName, source)
--- a/cmd/catalyst-node/catalyst-node.go
+++ b/cmd/catalyst-node/catalyst-node.go
@@ -284,6 +284,7 @@ func main() {
 	// Serf commands passed straight through to the agent
 	serfConfig := agent.Config{}
 	fs.StringVar(&serfConfig.BindAddr, "bind", "0.0.0.0:9935", "Address to bind network listeners to. To use an IPv6 address, specify [::1] or [::1]:7946.")
+	fs.StringVar(&serfConfig.AdvertiseAddr, "advertise", "0.0.0.0", "Address to advertise to the other cluster members")
 	fs.StringVar(&serfConfig.RPCAddr, "rpc-addr", "127.0.0.1:7373", "Address to bind the RPC listener.")
 	retryJoin := fs.String("retry-join", "", "An agent to join with. This flag be specified multiple times. Does not exit on failure like -join, used to retry until success.")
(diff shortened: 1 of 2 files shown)


cmd/catalyst-node/catalyst-node.go
```diff
@@ -29,35 +30,42 @@ import (
 
 const (
 	httpPort         = 8090
+	httpInternalPort = 8091
 	mistUtilLoadPort = 8042
 )
 
 var Version = "unknown"
+var serfClient *serfclient.RPCClient
```
leszko, line 38: I think this is ok as a global variable. We won't have multiple serf clients per one catalyst node. If you really don't want to have the global variable then you could always have something like `**serfclient.RPCClient` and pass it to `resolveNodeURL()`. But TBH I think it's not worth making it.

One thing that you could actually do is to add a wait with waitgroup before using the `serfClient`. Otherwise you may end up with `panic` if the HTTP endpoint is called before `serfClient` is initialised.

cmd/catalyst-node/catalyst-node.go
```diff
@@ -274,16 +283,19 @@ func main() {
...
 	// Catalyst web server
-	fs.StringVar(&cliFlags.HTTPAddress, "http-addr", fmt.Sprintf("127.0.0.1:%d", httpPort), "Address to bind for Catalyst HTTP commands")
+	fs.StringVar(&cliFlags.HTTPAddress, "http-addr", fmt.Sprintf("127.0.0.1:%d", httpPort), "Address to bind for external-facing Catalyst HTTP handling")
+	fs.StringVar(&cliFlags.HTTPInternalAddress, "http-internal-addr", fmt.Sprintf("127.0.0.1:%d", httpInternalPort), "Address to bind for internal privileged HTTP commands")
 
 	fs.StringVar(&config.serfRPCAddress, "serf-rpc-address", "127.0.0.1:7373", "Serf RPC address")
 	fs.StringVar(&config.serfRPCAuthKey, "serf-rpc-auth-key", "", "Serf RPC auth key")
 	serfTags := fs.String("serf-tags", "node=media", "Serf tags for Catalyst nodes")
 	fs.StringVar(&config.mistLoadBalancerEndpoint, "mist-load-balancer-endpoint", "http://127.0.0.1:8042/", "Mist util load endpoint")
+	fs.StringVar(&config.mistLoadBalancerTemplate, "mist-load-balancer-template", "http://%s:4242", "template for passing nodes to MistUtilBalancer")
```
leszko, line 293: Could you improve this flag's description? I think that from reading it, it's hard to grasp why you'd need it.

cmd/catalyst-node/catalyst-node.go
```diff
@@ -443,6 +456,12 @@ func startCatalystWebServer(httpAddr string, redirectPrefixes []string, nodeHost
 	glog.Fatal(http.ListenAndServe(httpAddr, nil))
 }
 
+func startInternalWebServer(internalAddr string) {
+	http.Handle("/STREAM_SOURCE", streamSourceHandler())
```
leszko, line 460: Shouldn't it be `/streamsource` (to follow the general rest api convention)?

cmd/catalyst-node/catalyst-node.go
```diff
@@ -478,11 +497,84 @@ func redirectHandler(redirectPrefixes []string, nodeHost string) http.Handler {
...
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		b, err := io.ReadAll(r.Body)
+		if err != nil {
+			glog.Errorf("error handling STREAM_SOURCE body=%s", err)
+			w.Write([]byte("push://"))
+			return
+		}
+		streamName := string(b)
+		glog.V(7).Infof("got mist STREAM_SOURCE request=%s", streamName)
+		dtscURL, err := queryMistForClosestNodeSource(streamName, "0", "0", "", true)
```
leszko, line 521: I'm not sure I understand why you query here Mist with lat/lon as `0/0`, could you maybe add some comment with the explanation?

cmd/catalyst-node/catalyst-node.go
```diff
@@ -478,11 +497,84 @@ func redirectHandler(redirectPrefixes []string, nodeHost string) http.Handler {
...
+	if err != nil {
+		return "", err
+	}
+	nodeName := u.Host
+	protocol := u.Scheme
+
+	member, err := getSerfMember(nodeName)
+	addr, has := member.Tags[protocol]
+	if !has {
+		glog.V(7).Infof("no tag found, not tag resolving protocol=%s nodeName=%s", protocol, nodeName)
```
leszko, line 550: Do we plan to tag all members with protocols?

cmd/catalyst-node/catalyst-node.go
```diff
@@ -478,11 +497,84 @@ func redirectHandler(redirectPrefixes []string, nodeHost string) http.Handler {
...
+		return streamURL, nil
+	}
+	u2, err := url.Parse(addr)
+	if err != nil {
+		glog.Errorf("node has unparsable tag!! nodeName=%s protocol=%s tag=%s", nodeName, protocol, addr)
```
leszko, line 555: Shouldn't we return here?

cmd/catalyst-node/catalyst-node.go
```diff
@@ -478,11 +497,84 @@ func redirectHandler(redirectPrefixes []string, nodeHost string) http.Handler {
 
 		rPath := fmt.Sprintf(pathTmpl, fullPlaybackID)
 		rURL := fmt.Sprintf("%s://%s%s", protocol(r), bestNode, rPath)
+		rURL, err = resolveNodeURL(rURL)
```
leszko, line 500: Why do we need to resolve the node here? Didn't we already query Mist for the best node and it's already resolved?

cmd/catalyst-node/catalyst-node.go
```diff
@@ -478,11 +497,84 @@ func redirectHandler(redirectPrefixes []string, nodeHost string) http.Handler {
...
+func querySerfForMember(name string) (*serfclient.Member, error) {
+	members, err := serfClient.MembersFiltered(map[string]string{}, "alive", name)
+	if err != nil {
+		return nil, err
+	}
+	if len(members) < 1 {
+		return nil, fmt.Errorf("could not find serf member name=%s", name)
+	}
+	if len(members) > 1 {
+		glog.Errorf("WARNING: found multiple serf members with the same name! name=%s count=%d", name, len(members))
```
leszko, line 571: nit: It's already an error message. No need to start it with `WARNING: `.

leszko (changes requested):

cmd/catalyst-node/catalyst-node.go
iameli: Yep!

cmd/catalyst-node/catalyst-node.go
iameli: MistUtilLoad assumes every node is accessible on 443, which isn't true in our staging or canary environments, so we can use Serf ot override here

PR #157, opened 2022-09-21
Makefile: add Dockerfile.local build

iameli:
I forgot about this trick to work around `.dockerignore` until now!

This will be useful for devs making changes to other projects locally but then testing with the containerized version.


hjpotter92 (commented):
lgtm

hjpotter92 (approved):

PR #163, opened 2022-09-23
Add USER_NEW trigger handler, split handlers code

### LINKED ISSUE #165: Add USER_NEW trigger handler
Add a webhook that is executed when Mist triggers `USER_NEW`.
Add example Mist configuration to route trigger to the correct endpoint.

red-0ne:
This PR adds a Mist triggers handler for the catalyst-node at `/triggers` endpoint it also intercepts `USER_NEW` trigger and parse its payload. No logic to control whether to allow or deny the user access to playback.

Refactor catalyst-node to extract endpoints to their own module.

How to test:
* Build the docker image with
```
make docker
```
* Run the following command to launch MistController:
```
docker run \
  --shm-size=1874m  --rm -it \
  -v $(pwd)/config:/config:rw \
  --hostname catalyst-one --name catalyst-one --network catalyst-nodes \
  -p 1935:1935 -p 8080:8080 -p 4242:4242 -p 7373:7373 \
  livepeer/catalyst:latest MistController -c /config/single-catalyst.json
```
* Send a stream to catalyst
```
ffmpeg -re -f lavfi -i testsrc=size=480x320:rate=30 -f lavfi -i sine -vf drawtext="text='%{localtime\:%X}':fontsize=20:fontcolor=white:x=7:y=7" -pix_fmt yuv420p -vcodec libx264 -vb 2000k -preset ultrafast -acodec aac -f flv 'rtmp://127.0.0.1:1935/live/stream+foo'
```

* Try to playback the stream on `http://localhost:4242/`
Note that we are using `config/single-catalyst.json` since it contains the `USER_NEW` trigger config.

Successful handling of the trigger will result in a working playback if `cmd/catalyst-node/handlers/triggers-handler.go`s `handleUserNew` returns `[]byte("true")` or no playback if `[]byte("false")` is returned.

Closes #165


leszko:
@red-0ne Before I review the code, could you please do the following points?
- Resolve conflicts with `main`
- Add some context in the description, a link to the GH Issue or Notion or something to understand the context of your change

Thanks!

cmd/catalyst-node/handlers/triggers-handler.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	streamName := lines[0]
+	connectionAddress := lines[1]
+	connectionID := lines[2]
+	connector := lines[3]
+	requestURL := lines[4]
+	sessionID := lines[5]
+
+	glog.Infof("streamName: %v, connectionAddress: %v, connectionId: %v, connector: %v, requestUrl: %v, sessionId: %v",
```
leszko, line 52: I see that all you do here is just to log the message. Then, what is the use of this trigger handler?

cmd/catalyst-node/handlers/redirect-handler.go
```diff
@@ -0,0 +1,260 @@
...
+	"sync"
+
+	serfclient "github.com/hashicorp/serf/client"
+	glog "github.com/magicsong/color-glog"
+)
+
+var mistUtilLoadPort = rand.Intn(10000) + 40000
+var getClosestNode = queryMistForClosestNode
+
+func RedirectHandler(redirectPrefixes []string, nodeHost string, serfClient *serfclient.RPCClient) http.Handler {
```
leszko, line 20: Good refactoring 🤝

cmd/catalyst-node/handlers/triggers-handler.go
```diff
@@ -0,0 +1,56 @@
...
+		payload, err := io.ReadAll(r.Body)
+
+		if err != nil {
+			w.Write([]byte("false"))
+			glog.Errorf("error %v", err)
+			return
+		}
+
+		triggerName := r.Header.Get("X-Trigger")
+
```
leszko, line 24: Is it possible to add a unit test?

leszko (changes requested):
@red-0ne I looked into your PR. The PR itself looks good (added a few minor comments). But... I still, even after reading your PR / GH Issue description cannot get the whole picture. 

Meaning:

1. Why do we need to handle this trigger? We don't do anything other than logging, so do we really need it?
2. When is the Mist trigger registered?
3. Why is this implemented in `catalyst` and not `catalyst-api` (like all other trigger handlers)?

red-0ne:
> 1. Why do we need to handle this trigger? We don't do anything other than logging, so do we really need it?

This handle will be used to implement gated playback. The current task is just to have the mechanism ready which means being able to instruct Mist to call our endpoint (specified in `config/single-catalyst.json`) when a `USER_NEW` event happens at Mist. Our handler has to catch the request and parses its payload. The response is for now hardcoded but the next task is to update it with signature verification logic which will decide whether the jwt token bearer is allowed to get payback from Mist or not.

> 2. When is the Mist trigger registered?

Mist is configured to call our endpoint in this file `config/single-catalyst.json`. Its content is missing newlines (MistController keeps writing it that way). We should commit a properly formatted version.

> 3. Why is this implemented in `catalyst` and not `catalyst-api` (like all other trigger handlers)?

Good question, which we asked with Yondon. We were mainly following this Access Control design https://www.notion.so/livepeer/c8c4ab6612144d8580668c0d65a2cfee?v=ebcb7856ce9a494484fed37699ec08e2 document (Lines 11 and 12)

cmd/catalyst-node/handlers/triggers-handler.go
red-0ne: The next task will be to use these info to decide whether to permit playback or not.

cmd/catalyst-node/handlers/triggers-handler.go
```diff
@@ -0,0 +1,56 @@
+package handlers
```
hjpotter92, line 1: nit: should the file be named `trigger-handler`?

config/single-catalyst.json
```diff
@@ -0,0 +1 @@
...ername":null},"debug":null,"defaultStream":null,"limits":null,"location":{"lat":0.0000000000,"lon":0.0000000000,"name":""},"prometheus":"koekjes","protocols":[{"connector":"AAC"},{"connector":"CMAF"},{"connector":"DTSC"},{"connector":"EBML"},{"connector":"FLV"},{"connector":"H264"},{"connector":"HDS"},{"connector":"HLS"},{"connector":"HTTP"},{"connector":"HTTPTS"},{"connector":"JSON"},{"connector":"MP3"},{"connector":"MP4"},{"connector":"OGG"},{"connector":"RTMP"},{"connector":"RTSP"},{"connector":"SDP"},{"connector":"SRT"},{"connector":"TSSRT"},{"connector":"WAV"},{"connector":"WebRTC"},{"connector":"livepeer-catalyst-node","redirect-prefixes":"stream","rpc-addr":"0.0.0.0:7373"}],"serverid":null,"sessionInputMode":"14","sessionOutputMode":"14","sessionStreamInfoMode":"1","sessionUnspecifiedMode":"0","sessionViewerMode":"14","sidMode":"0","triggers":{"USER_NEW":[{"default":"true","handler":"http://127.0.0.1:8090/triggers","streams":[],"sync":true}]},"trustedproxy":[]},"push_settings":{"maxspeed":null,"wait":null},"streams":{"stream":{"name":"stream","processes":[],"realtime":false,"source":"balance:http://localhost:8042/?fallback=push://","stop_sessions":false}},"ui_settings":null}
```
hjpotter92, line 1: is the file needed?

red-0ne: This file, as other config files are mainly used for local setup and exposed as examples. What bothers me is that MistController keep rewriting in a single line. Will update it to the adequate formatting.

hjpotter92: oh, ok. you can use `make dev` which creates a copy to `~/.config/livepeer/catalyst.json` and uses that for starting up `MistController`.

leszko:
Thanks for your answers @red-0ne . I think one important thing is that we should clarify if this change should be implemented in `livepeer/catalyst` or `/livepeer/catalyst-api`. For me, if there is no good reason to implement it in `livepeer/catalyst`, I'd implement it in `livepeer/catalyst-api`, because that's where we have all the trigger-handling logic.

@yondonfu wdyt?

iameli:
@leszko I steered @red-0ne toward implementing it in `catalyst-node` because we will need the `USER_NEW` trigger for other reasons in catalyst-node very soon — specifically viewer count enforcement will be a product of information that's only available to the multinode server.

iameli:
But in general I'd agree the the distinction between `catalyst-api` and `catalyst-node` is currently murky at best and we should either decide on a clear distinction between the two projects or merge them.

leszko:
> But in general I'd agree the the distinction between `catalyst-api` and `catalyst-node` is currently murky at best and we should either decide on a clear distinction between the two projects or merge them.

Yeah, totally 💯 Either make this distinction clear or merge the projects.

red-0ne:
Closed in favor of https://github.com/livepeer/catalyst/pull/188

PR #164, opened 2022-09-25
vmagent/victoriametrics: add environment variable config

iameli:
This allows `VM_` parameters to configure everything via environment variables, meaning we don't have to add all the VM options to livepeer-log.

### DIFF
--- a/cmd/catalyst-node/catalyst-node.go
+++ b/cmd/catalyst-node/catalyst-node.go
@@ -55,6 +55,8 @@ type catalystNodeCliFlags struct {
 	HTTPInternalAddress string
 	RedirectPrefixes    []string
 	NodeHost            string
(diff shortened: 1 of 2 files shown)


cmd/catalyst-node/catalyst-node.go
```diff
@@ -55,6 +55,8 @@ type catalystNodeCliFlags struct {
 	HTTPInternalAddress string
 	RedirectPrefixes    []string
 	NodeHost            string
+	NodeLatitude        float64
+	NodeLongitude       float64
```
hjpotter92, line 59: should we use `string` here, to avoid maybe floating arithmetic issues?

hjpotter92 (approved):
looks good, just worried about lat/long values being float

PR #170, opened 2022-09-27
catalyst-node: 301 redirect instead of 302

iameli:
An experiment to see if this leads to better properties when doing a changeover.


iameli:
nah, probably a bad idea

PR #173, opened 2022-09-29
Dockerfile: Create stages to strip binaries off debug symbols.

hjpotter92:
Reduces the final docker image size.


iameli (approved):
LGTM!

PR #177, opened 2022-09-29
catalyst-node: handle vod source urls without hitting load-balancer

emranemran:


yondonfu (approved):
LGTM!

Let's also add an issue to track using a shared def for stream prefixes so we don't need to manage stream prefixes here as well as in https://github.com/livepeer/catalyst-api/blob/main/config/stream_names.go#L12

yondonfu:
Looks like the Drone build is failing, but they've been failing for all PRs even before this one...Let's look into that separately.

emranemran:
> Looks like the Drone build is failing, but they've been failing for all PRs even before this one...Let's look into that separately.

File https://github.com/livepeer/catalyst/issues/178

emranemran:
> LGTM!
> 
> Let's also add an issue to track using a shared def for stream prefixes so we don't need to manage stream prefixes here as well as in https://github.com/livepeer/catalyst-api/blob/main/config/stream_names.go#L12

Filed https://github.com/livepeer/catalyst/issues/178

emranemran:
> Looks like the Drone build is failing, but they've been failing for all PRs even before this one...Let's look into that separately.

Yeah this one needs to be looked at - these failures have been consistent and fairly random for several weeks now.

PR #182, opened 2022-09-30
catalyst-node: properly redirect query parameters

iameli:
One more place where we need to redirect query parameters.


hjpotter92 (approved):
lgtm

PR #195, opened 2022-10-11
access-control: Fix stale cache test

victorges:
Turns out the stale cache refreshing in background was really not working, as the check done from inside the goroutine was actually checking if the entry was fully expired instead of checking if it is stale.


gioelecerati (approved):
<3 LGTM cc @red-0ne

PR #233, opened 2022-10-27
Fix E2E Tests

leszko:
- Fix TestMultiNodeCatalyst
- Add E2E into GH Action workflow
- Remove MistLoadTest, reasoning: - We should not do any load tests locally - These tests were too complex and flaky, therefore hard to maintain
 - Remove TestBootstrapNode, because the bootstrap feature was kind-of broken in ea28e6131c0be6c69ea2eb592fd3591457147f60, so we need to first thing how it should work


thomshutt (approved):

test/e2e/e2e_test.go
```diff
@@ -38,7 +35,7 @@ type cliParams struct {
 var params cliParams
 
 func init() {
-	flag.StringVar(&params.ImageName, "image", "livepeer/catalyst", "Docker image to use when loading container")
+	flag.StringVar(&params.ImageName, "image", "livepeerci/catalyst:main", "Docker image to use when loading container")
```
iameli, line 38: Hmmm, I think this will break the workflow of `make docker` then running the tests, because `make docker` currently builds `livepeer/catalyst`. Should either keep this as-is or update `Makefile` to make this new name

iameli (changes requested):
Mixed feelings about the removal of `catalyst-one.json` and `catalyst-two.json` — they're definitely hard to maintain but it's pretty useful to be able to spin up two nodes locally and play with them. Could we maybe expose the config file generation as a separate script so we can easily produce those two files on demand?

iameli:
(If that's hard, I'd be cool with merging as-is but keeping those two files around for a bit)

test/e2e/e2e_test.go
leszko: Fixed

leszko:
> (If that's hard, I'd be cool with merging as-is but keeping those two files around for a bit)

Ok, let's keep the `catalyst-one.json` and `catalyst-two.json` files for now.

leszko (commented):
@iameli I addressed your comments. PTAL.

.github/workflows/test.yaml
```diff
@@ -48,8 +48,14 @@ jobs:
       - name: misspell
         uses: reviewdog/action-misspell@v1
 
+      - name: Install FFMPEG
+        uses: FedericoCarboni/setup-ffmpeg@v1
+
+      - name: Build Docker image
+        run: make docker
```
iameli, line 55:
```suggestion
        run: make docker CATALYST_TAG=catalyst-e2e-test
```

.github/workflows/test.yaml
```diff
@@ -48,8 +48,14 @@ jobs:
...
+
       - name: Run tests with coverage
-        run: go test ./... --short --covermode=atomic --coverprofile=coverage.out
+        run: go test ./... --covermode=atomic --coverprofile=coverage.out
```
iameli, line 58:
```suggestion
        run: go test ./... --covermode=atomic --coverprofile=coverage.out --image=catalyst-e2e-test
```

iameli (changes requested):
One more tiny suggested change to make 100% sure we never run any e2e tests against something that wasn't just built. After that LGTM!

iameli (approved):
LGTM! (pending CI passing of course)

PR #254, opened 2022-11-08
.drone.yaml: better go test command

iameli:
Should fix e2e testing


hjpotter92:
not necessary anymore, as it seems to have landed in our main branch.

PR #257, opened 2022-11-09
Check for correct segmenting output paths in VOD E2E

thomshutt:
These have changed since the test was written, which is why it was failing.


iameli:
Yep! Tacked on https://github.com/livepeer/catalyst/pull/257/commits/6e0907fdd92eadd3cadd4182bebe8bedfae609b8 but it looks great

PR #293, opened 2022-11-21
e2e tests: refactor, add keyframe interval test

iameli:


config/catalyst-one.json
```diff
@@ -4,7 +4,16 @@
...
   },
-  "autopushes": null,
+  "autopushes": [
+    [
+      "stream+",
+      "/dump/$datetime_$wildcard.mkv?video=0,2,3,4,5"
+    ],
+    [
+      "videorec+",
+      "s3+https://GOOGI2ELGMPME3M4ROXG7WLY:PfdRoAD26Bvux%2BaRpsfKWVVC9ew80hcCGMDWotLQ@storage.googleapis.com/eli-recording-dump/$wildcard/$datetime/index.m3u8"
```
iameli, line 14: Already revoked this

config/catalyst-one.json
```diff
@@ -121,9 +129,118 @@
...
+      ],
+      "realtime": false,
+      "source": "push://",
+      "stop_sessions": false
+    },
+    "videorec": {
+      "name": "videorec",
+      "processes": [
+        {
+          "access_token": "2ea8c31f-8f14-46b2-a34b-ccb75b776926",
```
iameli, line 204: already rolled

cmd/catalyst-node/balancer/balancer.go
```diff
@@ -0,0 +1,297 @@
...
+	"net/url"
+	"os"
+	"os/exec"
+	"sync"
+
+	serfclient "github.com/hashicorp/serf/client"
+	glog "github.com/magicsong/color-glog"
+)
+
+type BalancerIface interface {
```
leszko, line 18: I think you can just call it `Balancer` and then the implementation `BalancerImpl`. I don't think Go has one convention for this, but the name of the interface is actually used everywhere, so that what should be shorter IMO.

cmd/catalyst-node/balancer/balancer.go
```diff
@@ -0,0 +1,297 @@
...
+	Start() error
+	UpdateMembers(members *[]serfclient.Member) error
+	Kill()
+	GetBestNode(redirectPrefixes []string, playbackID, lat, lon, fallbackPrefix string) (string, string, error)
+	QueryMistForClosestNodeSource(playbackID, lat, lon, prefix string, source bool) (string, error)
```
leszko, line 23: Maybe just `GetClosestNode`?

cmd/catalyst-node/balancer/balancer.go
```diff
@@ -0,0 +1,297 @@
...
+}
+
+type Balancer struct {
+	config   *Config
+	cmd      *exec.Cmd
+	endpoint string
+}
+
+// create a new load balancer instance
+func NewBalancer(config *Config) *Balancer {
```
leszko, line 39:
```suggestion
func NewBalancer(config *Config) *Balancer {
```

Nit: Comment not needed, does not explain anything.

cmd/catalyst-node/balancer/balancer.go
```diff
@@ -0,0 +1,297 @@
...
+		glog.Errorf("Error getting mist load balancer servers: %v\n", err)
+		return err
+	}
+
+	membersMap := make(map[string]bool)
+
+	for _, member := range *members {
+		memberHost := member.Name
+
+		// commented out as for now the load balancer does not return ports
```
leszko, line 65: Try to avoid commenting out code. I think you can just remove it.

cmd/catalyst-node/balancer/balancer.go
```diff
@@ -0,0 +1,297 @@
...
+	"os/exec"
+	"sync"
+
+	serfclient "github.com/hashicorp/serf/client"
+	glog "github.com/magicsong/color-glog"
+)
+
+type BalancerIface interface {
+	Start() error
+	UpdateMembers(members *[]serfclient.Member) error
```
leszko, line 20: You could change it to `UpdateMembers(members []string) error` or `UpdateMembers(members []Member`. Then, you won't have dependency on `serfclient` in the `balancer` package.

cmd/catalyst-node/catalyst-node.go
```diff
@@ -332,166 +135,86 @@ func main() {
...
-		return "", err
-	}
-
-	// Close the file
-	if err := tmpFile.Close(); err != nil {
-		return "", err
+	// Start main reconcillation loop - get members from serf and update MistUtilLoad
+	for {
+		members := <-memberChan
+		n.Balancer.UpdateMembers(members)
```
leszko, line 200: Not sure it's related to this PR, but we should start using `context` and `timeouts` for all network calls. E.g. here, if Mist does not reply or takes long to reply, it'll block the whole member update functionality. The same if when Serf does not reply.

cmd/catalyst-node/catalyst-node.go
```diff
@@ -332,166 +135,86 @@ func main() {
...
-	return tmpFile.Name(), err
 }
 
-func startCatalystWebServer(redirectPrefixes []string, httpAddr, nodeHost, gateURL string) {
-	http.Handle("/", redirectHandler(redirectPrefixes, nodeHost))
-	http.Handle("/triggers", accesscontrol.TriggerHandler(gateURL))
-	glog.Infof("HTTP server listening on %s", httpAddr)
-	glog.Fatal(http.ListenAndServe(httpAddr, nil))
+func (n *Node) startCatalystWebServer() {
+	http.Handle("/", n.redirectHandler(n.Config.RedirectPrefixes, n.Config.NodeHost))
```
leszko, line 205: I think you can move all this `redirectHandler` into the `balancer` package. It seems to be related only to the balancer functionality.

cmd/catalyst-node/catalyst-node.go
```diff
@@ -332,166 +135,86 @@ func main() {
...
+	glog.Infof("HTTP server listening on %s", n.Config.HTTPAddress)
+	glog.Fatal(http.ListenAndServe(n.Config.HTTPAddress, nil))
 }
 
-func startInternalWebServer(internalAddr string, lat, lon float64) {
-	http.Handle("/STREAM_SOURCE", streamSourceHandler(lat, lon))
-	glog.Infof("Internal HTTP server listening on %s", internalAddr)
-	glog.Fatal(http.ListenAndServe(internalAddr, nil))
+func (n *Node) startInternalWebServer() {
+	http.Handle("/STREAM_SOURCE", n.streamSourceHandler(n.Config.NodeLatitude, n.Config.NodeLongitude))
```
leszko, line 212: The same here, if the handler is only related to `balancer`, you could try to move it there.

cmd/catalyst-node/cluster/cluster.go
```diff
@@ -0,0 +1,239 @@
...
+	"time"
+
+	serfclient "github.com/hashicorp/serf/client"
+	"github.com/hashicorp/serf/cmd/serf/command/agent"
+	glog "github.com/magicsong/color-glog"
+	"github.com/mitchellh/cli"
+	"github.com/mitchellh/mapstructure"
+)
+
+type ClusterIface interface {
```
leszko, line 17: The same comment as in the case of `Balancer` => I think you can call it `ClusterManager` and the implementation `ClusterManagerImpl`.

leszko (commented):
Good PR. Thanks @iameli for this refactoring work. Added some comments.

Makefile
```diff
@@ -172,5 +172,10 @@ docker:
 docker-local:
 	tar ch ./bin Dockerfile.local | docker build -f Dockerfile.local -t "$(DOCKER_TAG)" --build-arg=GIT_VERSION=$(GIT_VERSION) --build-arg=BUILD_TARGET=full -
 
+.PHONY: test
 test: docker
 	go test ./test/e2e/*.go -v --logtostderr
+
+.PHONY: test-local
+test-local: docker-local
+	go test ./test/e2e/*.go -v --logtostderr
```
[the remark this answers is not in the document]
iameli: `test` runs `make docker` and `test-local` runs `make docker-local`. This is useful for testing new versions of binaries — for example you could make a change to your local `mistserver` instance, run `make mistserver test-local`, and run the e2e test suite that incorporates your changes.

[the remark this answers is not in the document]
iameli: (Arguably we could move `go test ./test/e2e/*.go -v --logtostderr` into its own action to avoid the duplication through)

cmd/catalyst-node/balancer/balancer.go
iameli: Ooh yeah that makes sense, will do

cmd/catalyst-node/balancer/balancer.go
iameli: True, though I think we'll need our own type for them very soon. Which I guess would be... `cluster.Node`?

iameli: Implemented in 508a8e2

PR #325, opened 2022-12-07
Add transcoding to E2E

leszko:
I think we should also get rid of `mist_config.go` and use the same sample configs we use for the local runs. But I see it as a separate PR.


thomshutt (approved):
Great that it ended up being a relatively small change, it's going to be really useful to have coverage of the transcode stage 🎉

PR #332, opened 2022-12-12
Manually bump catalyst-api version (to avoid Mist version bump)

thomshutt:
We're keeping Mist pegged to a version for now


hjpotter92:
i'd recommend updating the `release` value to a tag (or branch name) instead. commit value will get updated by the bot

Issue #373, opened 2023-01-23
Update Catalyst Docker image with w3 dependencies


leszko:

leszko:
We need to add the `w3` toolkit we will use into our Catalyst Docker image. 

This includes:
- Node.js
- All Node.js module dependencies
- `w3` CLI (or the script we'll use)

PR #396, opened 2023-02-07
Add livepeer-w3 to catalyst docker image

leszko:
We need a nodejs script in order to upload files into web3.storage.

It increases the Docker images size `2.22GB` => `2.46GB`

For the context, please check this PR: https://github.com/livepeer/go-tools/pull/17

fix https://github.com/livepeer/catalyst/issues/373

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -39,6 +39,13 @@ RUN	strip -s /opt/bin/*
 
 FROM	catalyst-${BUILD_TARGET}-build	as	catalyst-build
 
+# Install livepeer-w3 required to use web3.storage
+FROM	node:18.14.0 as node-build
+ARG	LIVEPEER_W3_VERSION=v0.2.2
+WORKDIR /app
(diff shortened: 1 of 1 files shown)


thomshutt (approved):

hjpotter92 (approved):
lgtm. one minor suggestion, if possible; could go-tools generate a package similar to how studio releases `livepeer-api` and `livepeer-www` binaries with node dependencies bundled? it'd then simply be a matter of adding a service in manifest which can use the `strip` build target from this docker manifest.

leszko:
> lgtm. one minor suggestion, if possible; could go-tools generate a package similar to how studio releases `livepeer-api` and `livepeer-www` binaries with node dependencies bundled? it'd then simply be a matter of adding a service in manifest which can use the `strip` build target from this docker manifest.

That would be nice. Though I'm not sure how much work it is to create releases and if it's worth it. For now, I've created a GH Issue: https://github.com/livepeer/go-tools/issues/18

PR #400, opened 2023-02-08
Add ipfs-car CLI

leszko:
Needed for the web3.storage integration

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -67,7 +67,7 @@ RUN	apt update && apt install -yqq \
 
 COPY --from=catalyst-build	/opt/bin/		/usr/local/bin/
 COPY --from=node-build		/app/go-tools/w3	/opt/local/lib/livepeer-w3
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -68,6 +68,7 @@ RUN	apt update && apt install -yqq \
 COPY --from=catalyst-build	/opt/bin/		/usr/local/bin/
 COPY --from=node-build		/app/go-tools/w3	/opt/local/lib/livepeer-w3
 RUN	ln -s /opt/local/lib/livepeer-w3/livepeer-w3.js /usr/local/bin/livepeer-w3
+RUN	npm install -g ipfs-car
```
hjpotter92, line 71: both these `RUN` statements should be merged, to have 1 less layer!

leszko: Right, updated.

hjpotter92 (approved):

PR #435, opened 2023-03-17
Small refactor to access-control

mjh1:
We need to make some changes to this to support accessKey gating (as well as jwt) so this is an inital refactor to tidy things up a bit to make the further changes easier.


leszko (approved):
LGTM

PR #437, opened 2023-03-20
Refactor access control to support forms of gating other than jwt

mjh1:
This refactor is for the access control work, it allows us to hook into the `GetPlaybackAccessControlInfo` function and pass our own request body in (for accessKey rather than jwt).

### DIFF
--- a/cmd/catalyst-node/handlers/access-control/access-control.go
+++ b/cmd/catalyst-node/handlers/access-control/access-control.go
@@ -5,6 +5,7 @@ import (
 	"encoding/base64"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"io"
 	"net/http"
 	"net/url"
@@ -30,9 +31,10 @@ type PlaybackAccessControlEntry struct {
 }
 
 type PlaybackAccessControlRequest struct {
-	Type   string `json:"type"`
-	Pub    string `json:"pub"`
-	Stream string `json:"stream"`
+	Type      string `json:"type"`
+	Pub       string `json:"pub"`
+	AccessKey string `json:"accessKey"`
+	Stream    string `json:"stream"`
 }
 
 type GateAPICaller interface {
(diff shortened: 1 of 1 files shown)


cmd/catalyst-node/handlers/access-control/access-control.go
```diff
@@ -134,36 +139,36 @@ func (ac *PlaybackAccessControl) handleUserNew(payload []byte) []byte {
...
+func (ac *PlaybackAccessControl) GetPlaybackAccessControlInfo(playbackID, cacheKey string, requestBody []byte) (bool, error) {
 	ac.mutex.RLock()
-	entry := ac.cache[playbackID][pubKey]
+	entry := ac.cache[playbackID][cacheKey]
 	ac.mutex.RUnlock()
 
 	if isExpired(entry) {
-		glog.Infof("Cache expired for playbackId=%v pubkey=%v", playbackID, pubKey)
-		err := ac.cachePlaybackAccessControlInfo(playbackID, pubKey)
+		glog.Infof("Cache expired for playbackId=%v pubkey=%v", playbackID, cacheKey)
```
victorges, line 148:
```suggestion
		glog.Infof("Cache expired for playbackId=%v cacheKey=%v", playbackID, cacheKey)
```

Similar logs below

cmd/catalyst-node/handlers/access-control/access-control.go
```diff
@@ -119,7 +118,13 @@ func (ac *PlaybackAccessControl) handleUserNew(payload []byte) []byte {
 		pubKey = claims.PublicKey
```
victorges, line 118: It would be even better to have this `jwt` logic (`pubKey` extraction) into a re-usable helper as well. Perhaps the helper gets the playback URL and it tells whether the playback is authorized or not. This would allow both assets and streams to have a consistent access control experience, being able to use jwt or webhooks on either.

This is not a blocker if we're running tight on schedule, but I think it would avoid creating a tech debt item which would be having livestream and VOD access control experiences completely different (instead of both types supporting both protocols).

WDYT? Does it involve a too big refactor to be more worth doing only as a follow up maybe next month?

victorges (approved):
LGTM with some nits!

cmd/catalyst-node/handlers/access-control/access-control.go
mjh1: i think that makes sense to me, so you're thinking this kind of thing: https://github.com/livepeer/catalyst/pull/437/commits/527686c30e4009de9ac07c5fb0a6a3ce2815b361

victorges: Yeah exatly! I just noticed 1 thing on that commit: you're using the `tkn` query-string for the access key, which is something that already exists on the mist side and AFAIK has a different meaning. We should avoid that potential conflict by using something else. @iameli should have more context of this.

I had suggested an explicit `accessKey` query-string instead, which also gains in readability IMO.

mjh1: @victorges could you give this another check if you're happy with the final version of this function. I've tested integrated it locally into catalyst-api and working 👍

cmd/catalyst-node/handlers/access-control/access-control.go
```diff
@@ -102,24 +105,19 @@ func (ac *PlaybackAccessControl) handleUserNew(payload []byte) []byte {
...
-		claims, err := decodeJwt(jwtToken)
-		if err != nil {
-			glog.Errorf("Unable to decode on incoming playbackId=%v jwt=%v", playbackID, jwtToken)
-			return []byte("false")
-		}
-
-		if playbackID != claims.Subject {
-
-			glog.Errorf("PlaybackId mismatch playbackId=%v != claimed=%v", playbackID, claims.Subject)
+		pubKey = extractKeyFromJwt(jwtToken, playbackID)
```
victorges, line 108: WDYT of just using `IsAuthorized` helper with the `requestURL` var as an arg? Then we get `accessKey` auth for livestreams as well! (it seems like we don't yet, right?)

cmd/catalyst-node/handlers/access-control/access-control.go
```diff
@@ -134,36 +132,73 @@ func (ac *PlaybackAccessControl) handleUserNew(payload []byte) []byte {
...
+		acReq.AccessKey = accessKey
+		cacheKey = accessKey
+	} else if jwt != "" {
+		acReq.Pub = extractKeyFromJwt(jwt, acReq.Stream)
+		if acReq.Pub == "" {
+			return false, fmt.Errorf("failed to extract key from jwt: %s", jwt)
+		}
+
+		acReq.Type = "jwt"
+		cacheKey = acReq.Pub
```
victorges, line 158: WDYT of adding a little prefix to these cache keys just to avoid potential conflicts? Specifically the `accessKey` value is 100% user-input, so they could end up providing something that conflicts with an existing pub key I guess.

```suggestion
		cacheKey = "accessKey_" + accessKey
	} else if jwt != "" {
		acReq.Pub = extractKeyFromJwt(jwt, acReq.Stream)
		if acReq.Pub == "" {
			return false, fmt.Errorf("failed to extract key from jwt: %s", jwt)
		}

		acReq.Type = "jwt"
		cacheKey = "jwtPubKey_" + acReq.Pub
```

victorges (approved):
Nice

cmd/catalyst-node/handlers/access-control/access-control.go
victorges: @mjh1 Checked the new version and looks good! Made some minor comments but overall already LGTM!

cmd/catalyst-node/handlers/access-control/access-control.go
mjh1: Ah sorry ha, I looked at it initially and just thought this was totally separate as it's dealing with some POSTed payload, but we get a url out of that payload so it's fine :) will fix that!

gioelecerati (approved):

cmd/catalyst-node/handlers/access-control/access-control.go
mjh1: @victorges what do you think about making playbackID a param for the `IsAuthorized` func? I'm just a bit concerned about assuming the url structure and extracting it from the path, seems a bit error prone and any time there's a new url format we'd have to adjust the function.

victorges: Sure I think that makes sense to me as well!

cmd/catalyst-node/handlers/access-control/access-control.go
victorges: Omg there would actually be a very real attack here: just send an `accessKey` for a resource that is gated via JWT, with the same value as the JWT public key, then you'd get access to the stream without a real JWT :o

mjh1: oh wow yeah good spot

PR #450, opened 2023-03-31
ci: try building platforms in parallel too

iameli:
This would unblock deploys for the common `-amd64` case while still eventually pushing a multi-arch `amd64/arm64` build by the end.


hjpotter92 (approved):

hjpotter92:
it's very weird that the `-debug` image took around 20 minutes, despite caches being available https://github.com/livepeer/catalyst/actions/runs/4576542110 whereas the stripped binary version was built faster despite it having a separate stage for stripping debug symbols.

PR #474, opened 2023-04-14
Update local catalyst setup

leszko:


leszko (commented):
Update catalyst dev configuration

leszko (commented):
Update local catalyst setup

config/catalyst-dev.json
```diff
@@ -99,7 +99,9 @@
         "connector": "livepeer-catalyst-api",
         "port": "7979",
         "source-output":"/tmp/livepeer/source",
-        "vod-pipeline-strategy":"catalyst"
+        "vod-pipeline-strategy":"catalyst",
+        "api-server":"http://127.0.0.1:3004",
```
iameli, line 103: Are you sure this doesn't break the non-Studio setup? We check this variable for whether we should enable mapic, setting up the PUSH_REWRITE trigger and whatnot https://github.com/livepeer/catalyst-api/blob/9b77aefdc7ec2c00a38f812c6024d2893400c773/config/cli.go#L76

leszko: Good point. I actually checked it and `make dev` worked fine :)

leszko: Should I check anything else than vod upload endpoint?

iameli: Try sending in a livestream, I assume it will break for lack of a Studio endpoint

iameli: Like push `rtmp://127.0.0.1/live/stream+foo` and playback `http://localhost:8080/hls/stream+foo/index.m3u8`

leszko: Push to `rtmp://127.0.0.1/live/stream+foo` doesn't work even before this PR. Does it work for you?

iameli: Ah my bad, didn't realize this config has "video" and not "stream" enabled. Works for me though with
```
docker run --rm -it -v (pwd)/config:/config:ro --hostname catalyst-one --name catalyst-one --shm-size=7553m -p 1935:1935 -p 8080:8080 livepeerci/catalyst:main MistController -c /config/catalyst-dev.json
```
and

```
ffmpeg -re -i ~/testvids/RocketLeague_2h_2sGOP_1080p60.mkv -c copy -f flv 'rtmp://localhost/live/video+eli'
```

leszko: @iameli I checked and the same command also works after my changes. Am I missing something?

PR #499, opened 2023-05-03
livepeer-log: handle log lines longer than 64k

iameli:
We were previously using `bufio.Scanner` that stops scanning when it hits a line too long. We don't want that. `bufio.Reader` does what we want.

### DIFF
--- a/cmd/livepeer-log/livepeer-log.go
+++ b/cmd/livepeer-log/livepeer-log.go
@@ -143,7 +143,7 @@ func main() {
 			panic(err)
 		}
 
-		print := func(text string) {
+		output := func(text string) {
 			fmt.Fprintf(os.Stderr, "INFO|%s|%d|||%s\n", procname, mypid, text)
 		}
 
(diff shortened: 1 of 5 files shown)


cmd/livepeer-log/livepeer-log.go
```diff
@@ -143,15 +143,27 @@ func main() {
...
+		}
+
 		// Actually print with lots of lines!
-		for _, pipe := range []io.ReadCloser{stderr, stdout} {
-			go func(pipe io.ReadCloser) {
-				scanner := bufio.NewScanner(pipe)
-				for scanner.Scan() {
-					text := scanner.Text()
-					fmt.Fprintf(os.Stderr, "INFO|%s|%d|||%s\n", procname, mypid, text)
+		for i, pipe := range []io.ReadCloser{stdout, stderr} {
```
thomshutt, line 151: Is it worth enforcing some sort of limit still, for when we accidentally dump a whole video file of binary data into the logs?

iameli: I dunno, feels like the wrong level to handle that kind of thing, considering this is a script that we'd like to get rid of completely after moving to Mist's first-class external service thing. I'd want Loki or maybe containerd to enforce some kind of rate limit cap probably though.

thomshutt (approved):

PR #501, opened 2023-05-03
ci: cross-compile livepeer-log and catalyst

iameli:
It's faster. [4m42s](https://github.com/livepeer/catalyst/actions/runs/4874482388/jobs/8695477165) down from [9m32s](https://github.com/livepeer/catalyst/actions/runs/4874239090).


iameli:
Tested the darwin-arm64 binaries and they worked just fine; I'm assuming that means the other ones worked fine too.

hjpotter92 (approved):

PR #518, opened 2023-05-22
[SRE-118] Dockerfile: Add gstreamer to base docker image with its plugins

hjpotter92:
Install gstreamer-tools and it's base/good/bad plugins.


iameli (approved):
LGTM!

PR #543, opened 2023-06-19
E2E: Highlight error detail msg

thomshutt:
We hit this failure again but didn't get any error detail 
image
so adding surrounding text as a sanity check

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -288,25 +288,7 @@ func requireReplicatedStream(t *testing.T, c *catalystContainer) {
 		glog.Info("Got HLS manifest!")
 		return true
 	}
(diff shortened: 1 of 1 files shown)


test/e2e/e2e_test.go
```diff
@@ -288,25 +288,7 @@ func requireReplicatedStream(t *testing.T, c *catalystContainer) {
 		glog.Info("Got HLS manifest!")
 		return true
 	}
-	require.Eventually(t, correctStream, 5*time.Minute, time.Second, errorMsg)
-}
-
-func requireNotReplicatedStream(t *testing.T, c *catalystContainer) {
```
thomshutt, line 294: Also noticed this method was unused

mjh1 (approved):

PR #573, opened 2023-07-15
livepeer in a box v3

iameli-streams:
Much better than before.


Makefile
```diff
@@ -165,11 +140,34 @@ docker:
...
+	-v $$(realpath config):/config:ro \
+	--rm \
+	-it \
+	--name catalyst \
+	--shm-size=4gb \
+	-p 1935:1935 \
+	-p 8888:8888 \
+	-p 4242:4242 \
+	-e ATHIEST=true livepeer/catalyst:latest MistController \
+	-c /config/full-stack.json
```
pwilczynskiclearcode, line 173: Maybe `box-dev` could be changed into a `docker-compose` service or coexist... like we have `dev-container` in infra repo

iameli: Maybe! Having RabbitMQ and CockroachDB as separate dependencies is certainly something I'm looking at to speed up the startup time. That said, I've always found docker-compose to be kind of a mediocre dev environment and I've never found myself using it for daily development more than a week or so. The cool thing about dev in a single container is that container can be seamlessly deployed to test with others!

But yeah, I might have to live with separate database/queue stuff for the sake of a quicker startup time. Rabbit takes soooo long.

pwilczynskiclearcode: I don't have that bad experience with docker-compose but I never used alternatives like `minikube`.

iameli-streams:
Closing in favor of https://github.com/livepeer/catalyst/pull/590

PR #630, opened 2023-08-23
test/e2e: implement record-tester CI against the box

iameli:
I think this is a super high-impact test so I've made a couple of sacrifices to get it in there. Mostly that means I'm using a static port-forward of `8888` instead of letting testcontainers-go pick the port to use. This will be fixed with https://linear.app/livepeer/issue/SRE-159/livepeer-in-a-box-allow-customization-of-hostnameip-on-boot, but in the meantime I've needed to [disable the parallel tests](https://github.com/livepeer/catalyst/pull/630/commits/48ab4222bcc22502e264318206e7e1f75828b4a1) in order to let them run properly.


test/e2e/box_record_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+	return catalyst
+}
+
+func startRecordTester(ctx context.Context) error {
+	err := run(
+		ctx,
+		"go",
+		"run",
+		"github.com/livepeer/stream-tester/cmd/recordtester",
```
leszko, line 92: Don't we want to run a dockerized record-tester? I think it would play better with all the test containers and networking.

leszko (approved):
Added one comment. Either way LGTM 👍

test/e2e/box_record_test.go
iameli: That would definitely be an improvement but there are lots of restrictions with the box right now where things are hardcoded to `127.0.0.1`, such as the redirect URLs from the catalyst-api load balancing. I tried and failed to get a containerized version working with host networking, resulting in [this bug](https://github.com/testcontainers/testcontainers-go/issues/1514). This localhost restriction will be removed with [SRE-159](https://linear.app/livepeer/issue/SRE-159/livepeer-in-a-box-allow-customization-of-hostnameip-on-boot) and we can revisit.

PR #655, opened 2023-09-12
box: document limitations; fixes for WebRTC

iameli:
-   Changed the hardcoded streams in the database snapshots to have
    easy-to-remember stream keys like `2222-2222-2222-2222`
-   Changed the built-in streams to use the H264ConstrainedHigh profile so there
    are no B-Frames in the output
-   Moved all references from `127.0.0.1` to `localhost`; this is needed for
    WebRTC/Coturn to work properly
-   Removed outdated references to `GOOS=linux` and `KILL=true`; these are the
    defaults now


iameli-streams:
-   Changed the hardcoded streams in the database snapshots to have
    easy-to-remember stream keys like `2222-2222-2222-2222`
-   Changed the built-in streams to use the H264ConstrainedHigh profile so there
    are no B-Frames in the output
-   Moved all references from `127.0.0.1` to `localhost`; this is needed for
    WebRTC/Coturn to work properly
-   Removed outdated references to `GOOS=linux` and `KILL=true`; these are the
    defaults now

Makefile
```diff
@@ -219,6 +219,7 @@ box-dev: scripts
 	-v $$(realpath config):/etc/livepeer:ro \
 	-v $$(realpath ./coredumps):$$(realpath ./coredumps) \
 	-e CORE_DUMP_DIR=$$(realpath ./coredumps) \
+	$(shell for line in $$(cat .env 2>/dev/null || echo ''); do printf -- "-e $$line "; done) \
```
pwilczynskiclearcode, line 222: Can we use `docker run --env-file=...` https://docs.docker.com/engine/reference/commandline/run/#env ?

docs/livepeer-in-a-box.md
```diff
@@ -1,10 +1,95 @@
...
+## Running the Box
+
+First, select a directory for persisting your database and video content; in
+this example we will be using `$HOME/livepeer-in-a-box`.
 
-## Dependencies
+```shell
+BOX_DIR="$HOME/livepeer-in-a-box"
+mkdir -p $BOX_DIR
+docker run \
```
pwilczynskiclearcode, line 39: Should we encourage to use one of `make` commands?

thomshutt (approved):

docs/livepeer-in-a-box.md
iameli: No, 'cause these are directions for someone that might be running it without developing it

PR #702, opened 2023-10-17
Add c2patool

leszko:


leszko:
> Note that this will get it in the Docker image but not in the tarball archive that will run during a deploy - that'll need to happen via something like [2f584db](https://github.com/livepeer/catalyst/commit/2f584db98524b577bd8a8449f836e0417c3ff06f).
> 
> That said, this is making me feel like we should probably just generate that archive off of the Docker image anyway for simplicity

Thanks for the comment. Right, generating the archive from Docker image would be great. 

Though, for this PR, I'll just add what you pointed out.

.github/workflows/build.yaml
```diff
@@ -77,6 +77,14 @@ jobs:
 
           cp ./scripts/livepeer-vmagent ./bin/livepeer-vmagent
 
+      - name: Build c2patool
+        run: |
+          apt update && apt install -y curl build-essential git
+          curl https://sh.rustup.rs -sSf | sh -s -- -y
+          source "$HOME/.cargo/env"
+          cargo install c2patool
```
iameli-streams, line 85: Probably this to match the other one?
```suggestion
          cargo install --version 0.6.2 c2patool
```

iameli-streams (approved):
Approved with tiny change

.github/workflows/build.yaml
pwilczynskiclearcode: https://github.com/marketplace/actions/cargo-install I find dedicated github actions working faster than manual installations. Maybe worth trying out

.github/workflows/build.yaml
```diff
@@ -77,6 +77,16 @@ jobs:
 
           cp ./scripts/livepeer-vmagent ./bin/livepeer-vmagent
```
pwilczynskiclearcode, line 79: https://github.com/actions-rs/toolchain
```
      - uses: actions-rs/toolchain@v1
        with:
          toolchain: stable
```
might be needed before using `cargo`

leszko: Thanks, trying it out!

pwilczynskiclearcode (approved):

leszko:
Thanks @pwilczynskiclearcode for the suggestion with the cargo GH Action!

PR #738, opened 2023-11-14
add procps

emranemran:
Mainly so that we upgrade the version of `top` running on our pods and get access to better info.


iameli (commented):
Sure but you can't rely on it being there any time soon, we don't roll pods much. What would you think of using this in catalyst-api instead? We're going to need something like it for the load balancing work. https://github.com/cloudfoundry/gosigar

emranemran:
> @emranemran Do we still need this with all the system monitoring stuff landing in catabalancer?

yes! this helps with per-process cpu/mem usage during specific operations in catalyst-api and the inter-parsing of these logs with catalyst-api helps narrow down issues quicker.

emranemran:
hmm do we think we don't need to reference it by kernel version? 
``` 

3.894 Reading state information...
3.913 3 packages can be upgraded. Run 'apt list --upgradable' to see them.
3.927 
3.928 WARNING: apt does not have a stable CLI interface. Use with caution in scripts.
3.928 
: Couldn't find any package by glob 'linux-tools-5.4.0-148-generic'

```

Dockerfile
```diff
@@ -80,6 +80,9 @@ RUN	apt update && apt install -yqq \
 	ffmpeg \
     	nodejs \
 	gstreamer1.0-tools gstreamer1.0-plugins-good gstreamer1.0-plugins-base gstreamer1.0-plugins-bad \
+	procps \
+	vnstat \
+	linux-tools-common linux-tools-generic linux-tools-5.4.0-148-generic \
```
hjpotter92, line 85: `linux-tools-generic` alone should be enough. it should pick up correct `linux-tools-<version>-generic` based on installed kernel version.

emranemran: Got it, updating.

iameli (approved):
sure

iameli:
Tests failing though, you. should rebase

PR #742, opened 2023-11-17
Speed up Catalyst build by not building c2patool from source

leszko:
The proper solution is to implement this: https://github.com/contentauth/c2patool/issues/149

In the meantime, however, since our build currently takes crazy ~1.5h, I added this hack to download c2patool from our own binaries instead of building it from sources. It's not too elegant, but speeds up our build CI significantly.


leszko:
I've seen the CI test fails, but I believe it's unrelated.

mjh1 (approved):

mjh1:
@iameli would you be able to use your admin powers to merge this? The test failure is the same as what we see on the main branch: https://github.com/livepeer/catalyst/actions/runs/6863479720

PR #761, opened 2023-12-06
Dockerfile: Generate a 100 years validity certificate for localhost

hjpotter92:
Will be mostly used for `MistOutWebRTC` connector.


pwilczynskiclearcode (approved):

iameli (changes requested):
Mmmmm I don't think I like this; if WebRTC implementations don't use perfect forward secrecy then downloading this certificate from the Dockerfile might allow attackers to decrypt passively-recorded WebRTC traffic.

Maybe WebRTC implementations do use forward secrecy and maybe there's other stuff happening but regardless generating the cert at runtime seems like a good idea to me

PR #778, opened 2023-12-13
ci: kick off update piplines when a binary build completes

iameli:
So that inplace deploys work properly.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -164,3 +164,6 @@ jobs:
           path: ${{ steps.branch-manifest.outputs.manifest-file }}
           destination: "build.livepeer.live/${{ github.event.repository.name }}/"
           parent: false
+
+      - name: Notify new build upload
+        run: curl -X POST https://holy-bread-207a.livepeer.workers.dev


hjpotter92 (approved):

.github/workflows/build.yaml
```diff
@@ -164,3 +164,6 @@ jobs:
           path: ${{ steps.branch-manifest.outputs.manifest-file }}
           destination: "build.livepeer.live/${{ github.event.repository.name }}/"
           parent: false
+
+      - name: Notify new build upload
+        run: curl -X POST https://holy-bread-207a.livepeer.workers.dev
```
pwilczynskiclearcode, line 169: It actually causes "empty" deployments docker images to staging while `archive` is not ready yet -

Wasn't the post archive notification enough?
https://github.com/livepeer/catalyst/blob/main/.github/workflows/build.yaml#L168-L169

iameli: I'm confused, you just linked the lines that I added and @hjpotter92 merged

The problem is that sometimes Docker finishes first and sometimes archive finishes first. So if we want both of those things to trigger an upgrade we need to trigger a bump-images run both times.

Or we could make a new CI action that runs after both are done, but that wouldn't 100% make sure the problem doesn't occur because other things can trigger holy-bread in between.

iameli: But I think it should be harmless, right? The worst part is just that it's sort of confusing in the middle there.

pwilczynskiclearcode: It already made one confusion https://discord.com/channels/423160867534929930/1006190767510917202/1184479881409593414

pwilczynskiclearcode: We could pass something like
``` -d '{"inputs":{"archives": {"main": "https://build.livepeer.live/catalyst/ef2fc1d.../livepeer-catalyst-linux-amd64.tar.gz"}}}'```
or in case of images
``` -d '{"inputs":{"images": {"livepeerci/api:master": "livepeerci/api:sha-26a57c2"}}}'```

https://docs.github.com/en/rest/actions/workflows?apiVersion=2022-11-28#create-a-workflow-dispatch-event--parameters

to limit what `update-values.py` tries to bump

pwilczynskiclearcode: Sth like https://github.com/livepeer/livepeer-infra/pull/1831/files

PR #796, opened 2023-12-18
mist version bump

pwilczynskiclearcode:


pwilczynskiclearcode:
It partially succeeded in https://drone.livepeer.fish/livepeer/mistserver/105
Just `docker-amd64-static-debug` failed. On the restart it succeeded https://drone.livepeer.fish/livepeer/mistserver/106 but other upload failed because of artefacts duplicates.

PR #812, opened 2024-01-11
Speed up Catalyst build by downloading c2patool binary

leszko:
Speed up Catalyst build by downloading c2patool binary (instead of building it from source)

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -85,7 +85,7 @@ jobs:
 
       - name: Download c2patool
         run: |
-          curl "https://storage.googleapis.com/build.livepeer.live/c2patool/0.6.2/c2patool-linux-${{ matrix.architecture }}" -o c2patool.tar.gz
+          curl "https://storage.googleapis.com/build.livepeer.live/c2patool/0.6.2/c2patool-{{ matrix.platform.name }}-${{ matrix.architecture }}" -o c2patool.tar.gz
           tar -xvzf c2patool.tar.gz -C ./bin/c2patool
 
       - name: Strip binaries of debug symbols
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -85,7 +85,7 @@ jobs:
 
       - name: Download c2patool
         run: |
-          curl "https://storage.googleapis.com/build.livepeer.live/c2patool/0.6.2/c2patool-{{ matrix.platform.name }}-${{ matrix.architecture }}" -o c2patool.tar.gz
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -83,16 +83,10 @@ jobs:
...
-        with:
-          crate: c2patool
-          version: "0.6.2"
-          cache-key: "${{ matrix.platform.name }}-${{ matrix.architecture }}"
-
-      - name: Copy c2patool
+      - name: Download c2patool
         run: |
-          cp "${HOME}/.cargo/bin/c2patool" ./bin/c2patool
+          curl "https://storage.googleapis.com/build.livepeer.live/c2patool/0.6.2/c2patool-${{ matrix.platform.name }}-${{ matrix.architecture }}.tar.gz" -o c2patool.tar.gz
```
pwilczynskiclearcode, line 88: You could try CDN-fied URL:
```
  curl "https://build.livepeer.live/c2patool/0.6.2/c2patool-${{ matrix.platform.name }}-${{ matrix.architecture }}.tar.gz" -o c2patool.tar.gz
```

leszko: Good idea, updated. Thanks Paweł!

.github/workflows/build.yaml
```diff
@@ -83,16 +83,10 @@ jobs:
       - name: Configure Rust Toolchain
         uses: dtolnay/rust-toolchain@stable
```
pwilczynskiclearcode, line 84: So do we still need rust-toolchain then?

leszko: Removed, thanks!

Dockerfile
```diff
@@ -11,8 +11,8 @@ ARG TARGETARCH
 RUN	apt update && apt install -yqq \
 	curl \
 	ca-certificates
-RUN curl https://storage.googleapis.com/build.livepeer.live/catalyst/772a57003e6d96c9a47ef18fccb51a0c61207074/livepeer-catalyst-linux-${TARGETARCH}.tar.gz -o /catalyst.tar.gz
-RUN tar xzf /catalyst.tar.gz
+RUN curl https://build.livepeer.live/c2patool/0.6.2/c2patool-linux-${TARGETARCH}.tar.gz -o /c2patool.tar.gz
+RUN tar xzf /c2patool.tar.gz
```
pwilczynskiclearcode, line 15: Having these two as a single `RUN` would reduce the size of the image:
```
RUN curl https://build.livepeer.live/c2patool/0.6.2/c2patool-linux-${TARGETARCH}.tar.gz -o /c2patool.tar.gz \
    && tar xzf /c2patool.tar.gz
```

pwilczynskiclearcode (approved):

PR #813, opened 2024-01-12
Include mp4box binary required by catalyst-api

leszko:
Related livepeer/catalyst-api PR: https://github.com/livepeer/catalyst-api/pull/1086


leszko:
Closing, because apparently the mp4box does not solve the root issue for all scenarios.

PR #925, opened 2024-04-12
[ENG-1889] manifest: Skip manifest updation for mistserver releases

hjpotter92:
MistServer would need to be manually updated.

### DIFF
--- a/manifest.yaml
+++ b/manifest.yaml
@@ -65,6 +65,7 @@ box:
       commit: 014094222fe2a3cbb8ebf2ca43c12546c73090cc
     release: catalyst
     skipGpg: true
+    skipManifestUpdate: true
     srcFilenames:
       darwin-amd64: livepeer-mistserver-darwin-amd64.tar.gz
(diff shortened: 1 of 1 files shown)


thomshutt (approved):

manifest.yaml
```diff
@@ -64,7 +64,7 @@ box:
       project: mistserver
       commit: 014094222fe2a3cbb8ebf2ca43c12546c73090cc
     release: catalyst
-    skipGpg: true
```
thomshutt, line 67: Do we not need to keep this too?

hjpotter92: i think we were publishing gpg signatures alongside our `livepeer/mistserver` builds. let me confirm once.

hjpotter92: ok, we aren't. force pushed another commit. i recall adding it to mist too; but can't remember when (or if it was in upstream :shrug: ).

leszko (approved):

PR #940, opened 2024-04-23
Dockefile - install procps and vnstast

pwilczynskiclearcode:


Dockerfile
```diff
@@ -110,6 +110,8 @@ RUN	apt update && apt install -yqq \
 	# for `shasum`
 	perl \
 	coturn \
+	procps \
```
hjpotter92, line 113: wrong target to install them in. should be instead at line 80.

PR #949, opened 2024-05-15
workflows: Disable builds for targets that aren't generally needed

hjpotter92:


thomshutt (approved):

.github/workflows/build.yaml
```diff
@@ -30,7 +30,7 @@ jobs:
           - amd64
         platform:
           - name: linux
-          - name: darwin
+          # - name: darwin
```
leszko, line 33: I suggest removing this. Commented-out code is usually confusing.

leszko (approved):

.github/workflows/build.yaml
```diff
@@ -26,11 +26,9 @@ jobs:
       fail-fast: false
       matrix:
         architecture:
-          - arm64
           - amd64
         platform:
```
pwilczynskiclearcode, line 30: Could be simplified to:
```yaml
platform:
  - linux
```
we don't use `name:` key now

pwilczynskiclearcode (approved):
It might be worth to have manual builds (workflow_dispatch), nightly build or simply a separate workflow for `arm64` as it's becoming more popular on clouds (and less importantly `darwin`). Just saying.

pwilczynskiclearcode:
@hjpotter92 Should we merge this change it?

PR #955, opened 2024-05-23
Fix livepeer-api build in Livepeer in a Box

leszko:


Makefile
```diff
@@ -10,6 +10,7 @@ DOCKER_TAG ?= livepeer/catalyst
 FROM_LOCAL_PARENT ?= scratch # for `make docker-local` and `make box-local`
 DOCKER_TARGET ?= catalyst
 BUILD_TARGET ?= full
+STUDIO_DIR=$(if $(wildcard ../studio/packages/www/static-build/.),"../studio/packages/api","../studio")
```
victorges, line 13: so this uses whole studio project if the static build exists or just the api in case it hasnt been built yet, is that it? i wonder if we wanna jut return an error with instructions on how to build www instead

victorges (approved):
lgtm

Makefile
leszko: I think the opposite:
- If `static-build` does not exists, then we need to build the whole studio project
- If `static-build` exists, then we can build only `api`

We could print instruction, but I think this is more convenient, because you don't need to read anything. The script handles it for you.

victorges: Ah ok then! LGTM too

